PR #6, opened 2015-06-12
Increased flexibility of NML2ChannelAnalysis

rgerkin:
This allows NML2ChannelAnalysis to be run more easily from within scripts, using keyword arguments.  
It also makes more extensive use of the -v (verbose) argument by applying it more globally.  
It also writes IVCurve files even in -nogui mode (since this should be independent of whether the IVCurve is displayed or not).  
It computes steady-state currents based on the end of the clampDuration, rather than the end of the entire simulation, removing the need for the duration to be artificially short.  
Lastly, it makes some edits to conform to style guidelines.  
This pull request is necessary for [this](https://github.com/VahidGh/ChannelWorm/blob/tests/tests/IVCurve_Model.ipynb) notebook to run correctly.


pgleeson:
Thanks @rgerkin.

PR #8, opened 2015-06-18
Refactored NML2ChannelAnalysis into components for integration with other modules (e.g. SciUnit)

rgerkin:


rgerkin:
The pull request passes travis for versions 2.7 and 3.4.  It fails 2.6 to my use of a dictionary comprehension.  I propose ignoring this and removing 2.6 testing from travis.  2.6 is obsolete and no one uses it anyway: http://astrofrog.github.io/blog/2015/05/09/2015-survey-results/

pgleeson:
@rgerkin maybe it's better to just skip that test in test.sh. There should be a TravicCI env var for the python version...

rgerkin:
So you are OK with having the Travis badge show failed, but you want the test.sh file to check the Python version before running?

pgleeson:
No, I was suggesting wrap the failing command in test.sh with something like:

    if [ "$TRAVIS_PYTHON_VERSION" == "2.6" ] ; then 
        pynml-channelanalysis NaConductance.channel.nml -nogui
    fi

so it runs all other tests in 2.6 still...

rgerkin:
@pgleeson OK, done.

pgleeson:
Thanks @rgerkin

PR #14, opened 2015-10-09
DEFAULTS is now a global dict for easier manipulation from outside scripts

rgerkin:


pgleeson:
Thanks @rgerkin. Have tested this but have some issues with the changed lines for the plotting. Will have to check through it again after SfN. 

Also, happy to have the code reformatted/standardised, but difficult to see the specific changes in a PR with updates & reformatting. Do you use any standard too for this? autopep8?

rgerkin:
@pgleeson Regarding the plotting, I thought it would be better for reproducibility to have the figures generated in a way that looks good in-line (i.e. in a notebook, rather than as independent windows.  If you have any suggestions for better visualization, I'd be happy to implement them.  

I don't use a tool for reformatting, mainly it's just things I catch as I'm looking the code and change manually.  The changes should be PEP-8 relevant, e.g. wrapping at 80 characters and making sure there is whitespace around any operators.  Sometimes I'll add or remove a line break for consistency with the rest of the file as well.

rgerkin:
@pgleeson 
What do you want me to do with this?

pgleeson:
Merged @rgerkin, thanks. Let me know if there are any other PRs or issues you'd like me to take another look at.

PR #27, opened 2016-07-19
Bug fixes and improvements that increase flexibility

rgerkin:


pgleeson:
Thanks @rgerkin, will have a look.

pyneuroml/analysis/NML2ChannelAnalysis.py
```diff
@@ -559,20 +561,23 @@ def make_iv_curve_fig(a,grid=True):
...
         % a.erev)
-    plt.xlabel('Membrane potential (V)')
-    plt.ylabel('Current (A)')
+    plt.xlabel('Membrane potential (mV)')
+    plt.ylabel('Current (pA)')
     plt.grid('on' if grid else 'off')
 
 
-def plot_iv_curve(a,hold_v,i,grid=True,same_fig=False,**plot_args):     
+def plot_iv_curve(a, hold_v, i, *plt_args, grid=True, same_fig=False,
```
pgleeson: This seems to be invalid?

Can you successfully run: **[./test.sh](https://github.com/NeuroML/pyNeuroML/blob/master/test.sh) -neuron** locally?

rgerkin: @pgleeson Yes (up until an unrelated key error), but that was in Python 3.  I see that the syntax I used above is only valid for Python 3 but not 2.  My goal here was to be able to pass in both positional and keyword arguments to `plot_iv_curve`, that could then be used to override any default options ultimately passed to `plt.plot`.  Keyword arguments suffice for almost everything, but positional ones are needed to override the third positional argument in plot, the string that you use to control linestyle and color, e.g. the 'ko-' in `plt.plot(hold_v, i, 'ko', ...)`

I am going to make this Python 2 compatible by doing this:  
```
def plot_iv_curve(a, hold_v, i, *plt_args, **plt_kwargs):     
  plt_kwargs['grid'] = plt_kwargs.get('grid',True)
  plt_kwargs['same_fig'] = plt_kwargs('same_fig',False) 
  ...
```
which achieves the same effect.

rgerkin: @pgleeson Fixed in [new PR](https://github.com/NeuroML/pyNeuroML/pull/29).

Issue #41, opened 2017-07-06
Making reloading of data more flexible


rgerkin:
For parallel processing I started executing in temporary directories (numbered by processor core, for example), which means that the location of the LEMS file (which is fixed) varies relative to the location of the file where I want the output to be written (which will be different for each core).  As I see it, the call to `reload_saved_data` as currently written (e.g. for [run with jNeuroML](https://github.com/NeuroML/pyNeuroML/blob/master/pyneuroml/pynml.py#L474) but it's the same for NEURON) requires a fixed relationship because the `relative_path` function assumes that `exec_in_dir`and `lems_file_name` will have a fixed relationship.  

I addressed this in uncommitted code by changing the signature of reload_saved_data to take the `exec_in_dir` path and the `lems_file` path as two separate arguments, and by making upstream functions responsible for providing a `lems_file` path that resolves to a file on disk that may or may not have anything to do with `exec_in_dir`.  

I want to check whether you have any reason to believe this will cause problems with other parts of the package or other use cases before I commit and pull request this.

pgleeson:
@rgerkin This sounds fine as long as everything else still works...

I had some changes in the development branch (still intend to do most work on master) which I've just merged, which relate to path handling. You'll want to test with these changes.

rgerkin:
Implemented in #43.

pgleeson:
Believe this is done...

PR #43, opened 2017-07-10
Separated exec_in_dir from directory with LEMS file

rgerkin:
`reload_saved_data` now passes these two arguments separately instead of combining them through the `relative_path` function.  I think this will require `lems_file_name` to always be resolvable to the LEMS file (either a full path or correct path relative to the working directory) independently of the value of `exec_in_dir`.  I've always been doing this anyway, so if any other part of the code that I am not aware of isn't, that would also need to be changed.  

This was motivated by #41, which can be closed when this is accepted.


pgleeson:
@rgerkin Thanks for this. It would be better though to still have the option of just specifying the lems file relative to the current dir as the only argument. What about having:
```
def reload_saved_data( lems_file_name,
                       base_dir='.', 
                       t_run=datetime(1900,1,1),
```
This would also have the advantage of not breaking existing code

rgerkin:
@pgleeson I've made the requested changes.

pgleeson:
Merged @rgerkin. Let me know if any other open issues of yours are proving bottlenecks...

rgerkin:
@pgleeson Thanks.  When will you push 0.3.1 to pypi?  I'm down to just a few libraries that I am still installing from dependency links on github.

pgleeson:
@rgerkin I've just put 0.3.1 on pypi. Let me know if you have any issues.

PR #48, opened 2017-09-25
Fixes issue 47

ccluri:
Fixes failure when there are comments inside STATE defined in a mod file. Uses regular expressions now.

Also a verification of chanToTest from SUFFIX keyword and an error for when not a distributed mechanism - defined as per 9.2.1 The NEURON block in 'The Neuron Book', 2006


pgleeson:
Thanks for this @ccluri, looks good.

PR #50, opened 2017-10-06
Added support for the new optional include path in jNeuroML

rgerkin:
And I used it successfully [here](https://github.com/openworm/CElegansNeuroML/blob/sciunit/CElegans/pythonScripts/c302/notebooks/RunMuscleModel-SciUnit.ipynb).  

I also used my freshly compiled `pyneuroml/lib/jNeuroML-0.8.3-jar-with-dependencies.jar` but did not commit that here to avoid pushing big binary files.  Can I assume that the one in the upstream pyneuroml repo is basically the same (i.e. that it supports the -I flag)?  

If this pull request is accepted I may ask for another pypi release...


pgleeson:
@rgerkin Ok, have merged this and updated the jar to the very latest.
Note I changed the parameter name 'include' to 'paths_to_include' as it was a bit ambiguous otherwise...

rgerkin:
Thanks!  Please post something when 0.3.2 is on pypi so I can update my dependencies to get it there.

pgleeson:
Latest is on pypi.

PR #52, opened 2017-11-05
Add execute_command_in_dir_with_realtime_output.

lungd:
Print output of command during execution.
That change is pretty helpful for Sibernetic simulations.


pgleeson:
@lungd Looks useful. I'll try it out...

slarson:
Hi!  If we can get this merged, we can point at it directly in the new OpenWorm dockerfile, which would be helpful for communicating the latest version to the community.  Any chance we can do this soon?  Thanks.

PR #53, opened 2018-02-22
New fail on missing includes feature requiring new PyLEMS version

rgerkin:
This PR changes two files.  The change in pynml.py adds several new options to the LEMS files reading function.  It requires the recent changes in PyLEMS that require PyLEMS 0.4.9.1 which is not yet on PyPI.  When it is on PyPI, I will update the setup.py here to remove the github dependency, and change the requirement for PyLEMS to >=0.4.9.1.


pgleeson:
@rgerkin All the changes you need for this should be on the master branch now.

rgerkin:
OK, I think I see them.

PR #57, opened 2019-01-22
Python 2 fix

rgerkin:
There are some circumstances under which `print_comment` will fail in Python 2 due to some of the inputs not having the correct encoding.  Adding `from __future__ import unicode_literals` solved this problem, and has no effect in Python 3.  

There are also many other commits here, but all of them seem to already be on your master branch, with you as the author (presumably you squashed them together and became the author, or rewrote them).


pgleeson:
Thanks @rgerkin . Testing these along with a few other updates here: https://github.com/NeuroML/pyNeuroML/tree/development

rgerkin:
Cool; I see it's already merged there.  Close this one?

pgleeson:
Thanks @rgerkin. I believe some of the changes you made in these commits had to be manually tested/added, but others like https://github.com/NeuroML/pyNeuroML/commit/84b370d5bf9bd71c876d3496be833970ba9a478d are showing up as yours...

PR #69, opened 2021-01-29
[pynml-summary] check for arguments

sanjayankur31:
Otherwise, it crashes and gives a trace.
Also adds a short usage snippet.

### DIFF
--- a/pyneuroml/pynml.py
+++ b/pyneuroml/pynml.py
@@ -17,7 +17,7 @@ import sys
 import subprocess
 import math
 from datetime import datetime
-import text
+import textwrap
 
 from pyneuroml import __version__
 from pyneuroml import JNEUROML_VERSION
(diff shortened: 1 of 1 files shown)


pyneuroml/pynml.py
```diff
@@ -17,6 +17,7 @@
 import subprocess
 import math
 from datetime import datetime
+import text
```
pgleeson, line 20: What module is text from? Is it py3.9?

sanjayankur31:
Eh! Oo, that's `textwrap`! Should be fine for 2.7+. (I can't find docs on
why using `text` worked here at the moment)

pyneuroml/pynml.py
sanjayankur31: Corrected this now.

PR #75, opened 2021-02-10
Improve graph options

sanjayankur31:


sanjayankur31:
@pgleeson : I've limited "min" for minus/negative levels. Using "m" makes it rather complex since that could also be a GraphViz engine in the future.

pgleeson (approved):

PR #79, opened 2021-02-17
Add issue templates

sanjayankur31:
Add github issue templates


pgleeson:
Good idea... Maybe add for NeuroML2 repo also, customised for feature requests for the standard (i.e. what simulator you need to use it with)

PR #82, opened 2021-02-26
doc: document usage of generate_plots utility function

### LINKED ISSUE #70: Add method documentation for `generate_plots`
It's a very very useful and so commonly used helper method, so we should have method documentation for it that includes an example or two.

sanjayankur31:
Fixes #70


pgleeson:
Looks good!

PR #83, opened 2021-02-26
enh(ci): enable GH actions

sanjayankur31:
This should enable GitHub actions here for CI.

We can also set up an action to automatically publish releases/tags to PyPi. I'll look into that next.


sanjayankur31:
Merging to see if it works.

PR #85, opened 2021-03-01
docs: add information about necessary dependencies

sanjayankur31:


sanjayankur31:
@pgleeson : since I'm not on a Ubuntu based distro: could you check the command I've included to see if it's correct please?

PR #107, opened 2021-06-18
Add cell builder utility functions

sanjayankur31:


sanjayankur31:
Not ready for review yet, got to move some of these to libNeuroML. Will do that after the current PR there has been merged, since this will require regenerating nml.py again

sanjayankur31:
I'll rebase this on development once #121 is merged to resolve the conflicts etc.

sanjayankur31:
@anujanegi @shayan823 since you're working with detailed cell models, could you have a look at these to see if they would be helpful? If you can think of any other improvements that we can make to these too, please do note them :)

sanjayankur31:
@pgleeson : as you noted in #170, I'm now wondering: would these fit better in libNeuroML? We added a few helper functions there already to support these here (to get segment groups, and we've got the `add` etc. helper methods there too). 

I guess strictly speaking libNeuroML is for reading/writing NeuroML, and we add a few helpers there to make writing easier. So, it's fine for these to be here because they're much more advanced functions. 

We should decide now because moving these to libNeuroML from pyNeuroML later will become a breaking change, although we could just keep wrappers in pyNeuroML to keep the API unchanged to make it a backwards compatible change instead.

pyneuroml/utils/CellBuilder.py
```diff
@@ -0,0 +1,296 @@
...
+    :type spec_cap: str
+    :param group: segment group to modify
+    :type group: str
+    """
+    cell.biophysical_properties.membrane_properties.specific_capacitances.append(
+        SpecificCapacitance(value=spec_cap, segment_groups=group)
+    )
+
+
+def add_channel_density(
```
anujanegi, line 251: How about option for [ChannelDensityNernst](http://neuroml.github.io/org.neuroml.model/org/neuroml/model/ChannelDensityNernst.html) here? @sanjayankur31

anujanegi:
Suggestion: support for setting [species](http://neuroml.github.io/org.neuroml.model/org/neuroml/model/Species.html) for cell.biophysical_properties.intracellular_properties.species?

pyneuroml/utils/CellBuilder.py
sanjayankur31: hrm, the channeldensity variants look quite specific. I guess we could look into extending the method to allow them all to be added. Let me think about that :+1:

anujanegi: Maybe! It's something I use, hence noticed it instantly.

sanjayankur31:
@pgleeson I think this is now ready for review. I pulled in changes from the experimental branch too---it's just a refactored version of those now.

shayan823:
We can try adding a function that creates a new segment group i.e. a subset of the existing segment groups and eventually adding it to the morphology of the cell. It can have utilities like adding inhomogeneous parameters as well.

pyneuroml/utils/CellBuilder.py
```diff
@@ -0,0 +1,296 @@
...
+    """Set the resistivity of the cell
+
+    :param cell: cell to modfify
+    :param resistivity: value resistivity to set with units
+    :type resistivity: str
+    :param group: segment group to modify
+    :type group: str
+    """
+    cell.biophysical_properties.intracellular_properties.resistivities = [
+        Resistivity(value=resistivity, segment_groups=group)
```
shayan823, line 243: Setting multiple resistivity values for different segment groups is not possible. Only the latest updated resistivity gets reflected in the NeuroML code. It can be rectified by using what I tried locally:

cell.biophysical_properties.intracellular_properties.resistivities.append(Resistivity(value=resistivity, segment_groups=group)

sanjayankur31: Thanks, that's a bug. We should be appending here instead of overwriting the list.

pyneuroml/utils/CellBuilder.py
```diff
@@ -0,0 +1,353 @@
...
+    :type spec_cap: str
+    :param group: segment group to modify
+    :type group: str
+    """
+    cell.biophysical_properties.membrane_properties.specific_capacitances.append(
+        SpecificCapacitance(value=spec_cap, segment_groups=group)
+    )
+
+
+def add_channel_density_v(
```
sanjayankur31, line 271: How does this look @anujanegi ?

Is this fine, or do you think specific helper functions for each type of ChennelDensity.. would be better?

Note that this uses the `add()` function for MembraneProperties. You'll need to use the libNeuroML from this PR for that:

https://github.com/NeuralEnsemble/libNeuroML/pull/127

(the schema had a bug that we corrected, and so libNeuroML had to be regenerated).

sanjayankur31:
> Suggestion: support for setting [species](http://neuroml.github.io/org.neuroml.model/org/neuroml/model/Species.html) for cell.biophysical_properties.intracellular_properties.species?

I've added a generic functions to add any intracellular/membrane properties now. How do these look?

sanjayankur31:
Note: needs https://github.com/NeuralEnsemble/libNeuroML/pull/127 to be merged first, since that includes updates to the libNeuroML python API that are required here.

PR #112, opened 2021-07-05
Consolidate logging (and other minor tweaks)

sanjayankur31:
- uses logging where appropriate,
- removes print_comment_v, and makes print_comment the default pretty printed comment function. It'll *always* print. For debuginfo and warnings and so on, we use the loggers.


sanjayankur31:
Refactoring, mostly with cosmetic changes. All tests pass, so confirmation that this hasn't broken any functionality. Merging (so I can open more PRs, which will otherwise conflict with this one).

pgleeson:
Thanks @sanjayankur31 . From a qucik test I see that the output from jnml, which had been displayed by default when run in pynml does not show any more. This is useful debugging info. Can you change that?

sanjayankur31:
Sure---since it is debugging info, it can be seen when `-v INFO` or `-v DEBUG` are provided. Do we always want this info to be always shown to users?

sanjayankur31:
If a command fails, it will be printed: these are at the `CRITICAL` level. For example:  https://github.com/NeuroML/pyNeuroML/blob/development/pyneuroml/pynml.py#L1649

PR #114, opened 2021-07-06
Create python-publish.yml

sanjayankur31:
Sets up the default GH action workflow to automatically publish releases on GH to PyPI


sanjayankur31:
Note: requires setting up the PyPi authentication token as a secret on GH.

sanjayankur31:
> @sanjayankur31 Does something still need to be done with secrets before this gets merged?

The PyPI API token needs to be stored as a "secret" on GitHub so that it can access it from the action:

https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#publishing-to-package-registries

PR #116, opened 2021-07-19
Finish consolidating logging

sanjayankur31:
We now use logger.info to display information to the user. The default logger level is also INFO, so all of these messages are displayed by default. Information that should not be shown to the user by default should use logger.debug.


pgleeson (approved):
Looks good.

PR #124, opened 2021-09-24
feat: format with black

sanjayankur31:
purely cosmetic, just ran black over the code to format it.


sanjayankur31:
Tests pass, merging (so that I can work on other PRs which will otherwise conflict with this one :()

PR #143, opened 2021-11-25
Feat/extra requires

sanjayankur31:


sanjayankur31:
Another py2 failure :(

sanjayankur31:
failing because we haven't created a `development` branch in pyelectro and neurotune, but our CI is set to install all our deps from the development branches. So just creating these in those repos will fix the tests.

PR #152, opened 2021-12-02
NeuroML 2.2 release

pgleeson:


sanjayankur31:
Something is pulling in `typing` which isn't needed on 3.7+, and that seems to be causing the error:

https://github.com/python/typing/issues/573

It doesn't seem to be any of our tools---we have `typing; python_version<"3.5"` everywhere I can find, so could be one of the deps  maybe..

pgleeson:
@sanjayankur31 It's probably because this test on master uses the master branch of libneuroml pylems, etc. which haven't been updated in a while, which probably have previous typing/py39 issues. Should be resolved when those repos have dev merged to master (next week...)

PR #155, opened 2021-12-07
Feat/increase tuning verbosity

sanjayankur31:


sanjayankur31:
py2 build will begin to pass once https://github.com/aarongarrett/inspyred/pull/24 is merged

PR #156, opened 2021-12-08
pip install modelspec version for dev

pgleeson:


sanjayankur31:
@pgleeson installing neuromllite should correctly pull in modelspec now instead of requiring it to already be installed, so we may not need this one now?

pgleeson:
This has other fixes too...

pgleeson:
Mainly it only test neuron related examples (incl tuning) on py3, so the py2 tests pass...

PR #158, opened 2021-12-09
Specifying universal_newlines so "realtime output" works correctly

mwatts15:
- Also, printing stderr output from jnml command

### DIFF
--- a/pyneuroml/pynml.py
+++ b/pyneuroml/pynml.py
@@ -36,7 +36,6 @@ except ImportError:
 
 import matplotlib
 import lems.model.model as lems_model
-from lems.model.fundamental import Include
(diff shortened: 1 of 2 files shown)


sanjayankur31:
Thanks @mwatts15 . I think this looks good.  I see that the two realtime functions are marked as needing refactoring but we can do that later. 

For this PR, would be able to add a quick test similar to the one for the non-realtime command runner please? Hopefully, copying it and making a few minor tweaks should be sufficient.

https://github.com/NeuroML/pyNeuroML/blob/development/tests/test_pynml.py#L101

sanjayankur31 (approved):
thanks, that looks good. Merging.

PR #160, opened 2021-12-13
Add swc export to pynml

sanjayankur31:
We should ideally have some unittests for the swc-export too

- Also regenerates the man pages


sanjayankur31:
this is resulting in some cyclic imports so will need some refactoring, so let's leave it for this release. Marking as draft instead.

PR #172, opened 2022-06-22
To v0.7.0

pgleeson:
Drops support for python 2
Adds initial support for running with EDEN: `pynml LEMS_model.xml -eden`
Fixes realtime output when running external processes: #171 
Minor updates to PovRay export

PR #173, opened 2022-06-26
Update ChannelInfo_TEMPLATE.md

shayan823:
Text overflowing in the first column, so adding a max-width to it and adding $channel.notes into <div> to avoid rendering issues.

### DIFF
--- a/pyneuroml/analysis/ChannelInfo_TEMPLATE.md
+++ b/pyneuroml/analysis/ChannelInfo_TEMPLATE.md
@@ -6,12 +6,14 @@ Channel information
 <table>
 #foreach ($channel in $channels)##
     <tr>
-<td width="120px">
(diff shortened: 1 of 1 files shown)


sanjayankur31 (approved):

sanjayankur31:
lgtm, will merge once CI finishes.

sanjayankur31:
Hrm, @shayan823: could you rebase this against development please? Just noticed it's against master. There should be no py2.7 in development too since we dropped support there IIRC.

shayan823:
Okay, then I should close this PR and instead open a new one against development? And are you saying that for further commits I should consider python 3?

sanjayankur31:
Yeh, we dropped py2 support, so only focus on py3. All PRs are made against development, and we merge development into master when we cut a new release. See #172 for example

PR #174, opened 2022-06-28
update changes to the template file

shayan823:


sanjayankur31:
So we're just making sure that the background for the channel notes is white---was this causing an issue (with GitHub dark mode?)

shayan823:
I tried adding the channel notes in a <div> tag as well. Actually, the line breaks in the notes of the neuroml file are not allowing the images to render properly. 
When tried adding max-width=120px to the <td> it wasnt helping so I reverted it back. 

Currently, the images are rendered properly except that the text in the first column is wrapped. Have a look at [this](https://github.com/OpenSourceBrain/BahlEtAl2012_ReducedL5PyrCell/tree/issue_12_test/NeuroML2/channel_summary).

pgleeson:
@shayan823 Just to clarify, are you saying this is a fix already or still being tested?

shayan823:
It's still being tested. As its fails to render the image when we have line breaks in the notes. 
We can parse the channel NeuroML files such that notes have no line breaks keeping the template as it is now.

pgleeson:
FYI the use of `<sup>` was a quick and dirty hack to allow small text for the longish descriptions. There is certainly a better way to do that which may give less problems...

PR #176, opened 2022-08-05
Feat/plotly plotting

sanjayankur31:


pyneuroml/plot/Plot.py
```diff
@@ -0,0 +1,430 @@
...
+        plt.savefig(save_figure_to, bbox_inches="tight")
+        logger.info("Saved image to %s of plot: %s" % (save_figure_to, title))
+
+    if show_plot_already:
+        plt.show()
+
+    return ax
+
+
+def generate_interactive_plot(
```
sanjayankur31, line 231: @irahulsonkar @anujanegi @shayan823 : this is the new function (see the test file for example usage too).

PR #177, opened 2022-08-08
Update channel export

pgleeson:


pgleeson:
Thanks for this @shayan823. I tweaked it a bit to allow an expandable section for the notes, so it's a bit more compact, e.g. https://github.com/OpenSourceBrain/SmithEtAl2013-L23DendriticSpikes/tree/master/NeuroML2/channel_summary

PR #182, opened 2022-08-10
For 2d plots

pgleeson:


pyneuroml/utils/__init__.py
```diff
@@ -0,0 +1,87 @@
+import neuroml
+
+def extract_position_info(nml_model, verbose):
```
sanjayankur31, line 3: Is this something users can use too? Otherwise we may want to make it private by naming it `_extract_position_info` (leading underscore?)

Also thinking if it should go into pyneuroml/plot/__init__.py instead since it's sort of specific to the plotting bits and not a general util? :thinking:

pgleeson: I mainly pulled it out so it could be used in the 3D plot too to support networks also. It's based on what's in the povray export...

Yes, it's mainly used for plotting now, but could be used for other forms of analysis, as a general purpose way to get quick dicts of positions (and connections/inputs, etc eventually), so maybe leave here for now...

sanjayankur31: Sounds good. Will document it at some point and try to add a test too.

sanjayankur31:
This looks fine now @pgleeson : OK to merge to dev?

pgleeson:
Yep, good with me to merge this to dev..

pgleeson:
Woo!
Screenshot 2022-08-12 at 16 38 13

sanjayankur31:
Nice! Merging this now.

PR #186, opened 2022-08-31
Use development branch of libNML in requirements-dev

pgleeson:
Needed so that nmllite tests pass. Defaulting to master libnml means a version less than the current required version for nmllite is used.

libNML dev will always be the latest and best tested, so probably not an issue when a dev install of pynml is required

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,7 +7,7 @@ NEURON
 ppft
 
 # change these to development if working on the development branch
-git+https://github.com/NeuralEnsemble/libNeuroML.git@master#egg=libNeuroML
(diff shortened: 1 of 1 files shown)


sanjayankur31:
I think we replace "master" with "development" (or "experimental") in the CI when working on the develop/experimental branches. Maybe we haven't done that for nmllite yet?

sanjayankur31:
-> https://github.com/NeuroML/pyNeuroML/blob/master/.github/workflows/ci.yml#L31

pgleeson:
Closing as versions are all managed in NeuroMLlite GHA scripts now

PR #193, opened 2022-10-14
Move mod2nml script to neuron utilities

sanjayankur31:
WIP: need to add some tests


sanjayankur31:
@shayan823: since you've worked with this one quite a bit---how does this look? It just includes it into the pyneuroml.neuron module basically, doesn't change the logic at all.

@pgleeson : should be ready to merge already---didn't change the logic of the functions at all.

sanjayankur31:
Docs will look like this:

https://pyneuroml--193.org.readthedocs.build/en/193/pyneuroml.neuron.html#pyneuroml.neuron.export_mod_to_neuroml2

sanjayankur31:
We can include this in the release too @pgleeson : doesn't modify any older code, just refactors it to include the mod2nml function in the `pyneuroml.neuron` sub module

pgleeson:
Sure, thanks!

PR #216, opened 2023-03-10
pynml-modchananalysis excepting on division by zero

borismarin:
Fixes #202, #165

Issue #165, opened 2022-04-12
Add a check in pynml-modchananalysis to see whether mod files are compiled yet


pgleeson:
Running `pynml-modchananalysis my_mod` will result in an obscure error if the mod files haven't been compiled in that dir. It should either:
- give a more detailed warning
- assume it should compile the mod files and load the appropriate library and proceed (if possible in python)

PR #217, opened 2023-03-20
test(modchananalysis): test to check that nrnivmodl is run when required

sanjayankur31:
See #165
See #216

### DIFF
--- a/test-ghactions.sh
+++ b/test-ghactions.sh
@@ -143,7 +143,7 @@ if [ "$run_neuron_examples" == true ]; then
     echo "################################################"
     echo "##   Test analysis of channel in mod file"
 
-        nrnivmodl
+        # do not run nrnivmodl, modchananalysis should run it if required
         pynml-modchananalysis -stepV 20  NaConductance  -dt 0.01 -nogui
 
 


borismarin (commented):
Great. I have also tested (locally) the case when a [nrnivmodl fails to compile](https://github.com/NeuroML/pyNeuroML/blob/b162b61655070c62f532865ab677893739e66b1d/pyneuroml/neuron/analysis/HHanalyse.py#L208) (by introducing a syntax error into the mod file by hand), in which case we return 1 (maybe not the best error handling) - should our tests catch that as well?

borismarin:
I am also foreseeing trouble trouble with the [hardcoded path for the compiled mechs](https://github.com/NeuroML/pyNeuroML/blob/b162b61655070c62f532865ab677893739e66b1d/pyneuroml/neuron/analysis/HHanalyse.py#L214). Seems to work fine for my machine (Linux x86_64) as well as the CI virtual machine - but I do not know what neuron does nowadays for other architectures. We could try and build the path using `platform.processor()`, but I am not sure nrnivmodl will always agree to that. Maybe we could have someone with a fancy new Apple silicon test it?

sanjayankur31:
> Great. I have also tested (locally) the case when [nrnivmodl fails to compile](https://github.com/NeuroML/pyNeuroML/blob/b162b61655070c62f532865ab677893739e66b1d/pyneuroml/neuron/analysis/HHanalyse.py#L208) (by introducing a syntax error into the mod file by hand), in which case we return 1 (maybe not the best error handling) - should our tests catch that as well?

Yeh, that'll be good to have as well---the more tests, the better! 



> I am also foreseeing trouble with the [hardcoded path for the compiled mechs](https://github.com/NeuroML/pyNeuroML/blob/b162b61655070c62f532865ab677893739e66b1d/pyneuroml/neuron/analysis/HHanalyse.py#L214). Seems to work fine for my machine (Linux x86_64) as well as the CI virtual machine - but I do not know what neuron does nowadays for other architectures. We could try and build the path using `platform.processor()`, but I am not sure nrnivmodl will always agree to that. Maybe we could have someone with a fancy new Apple silicon test it?

Ah yeh. I think there are some bits in org.neuroml.export etc too that also need to be updated for this. `nrnivmodl` uses `@host_cpu@` which is set to `${CMAKE_SYSTEM_PROCESSOR}`:

https://github.com/neuronsimulator/nrn/blob/master/bin/nrnivmodl.in#L4

https://github.com/neuronsimulator/nrn/blob/master/cmake/ConfigFileSetting.cmake#L36


https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PROCESSOR.html?highlight=cmake_system_processor

https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR

I think `platform.processor` should be find to use---it's used in the `platform.uname` output, which we'd expect to be similar/same as `uname`?

I think @pgleeson did just recently get a fancy new macbook to test this out for us :P

sanjayankur31:
I'll merge this and we can make more tweaks to the original PR, and continue the discussion etc. there too.

PR #218, opened 2023-04-19
feat: add method to rotate cell

sanjayankur31:
Question: should this go into libneuroml as a helper function to the `Cell` class?

TODO: a good test would be really nice. At the moment, I check with visual inspection to see that the result is fine, but that's not really a good test.


sanjayankur31:
Depends on https://github.com/NeuralEnsemble/libNeuroML/pull/160

PR #235, opened 2023-07-04
feat(generate-if-curve): allow specifying segment and fraction along

sanjayankur31:
To allow analysis of multi-compartmental cells with inputs at different locations


pgleeson:
@sanjayankur31 Maybe they should be segid=0 and fract=0.5 by default instead of None?

sanjayankur31:
We decided to leave it as None since non compartmental cell models would not take these values. The schema docs say that 0 and 0.5 will be used if these fields are left empty. Merging this now.

PR #238, opened 2023-07-07
Feat: extend morphology plotting methods to accept a file path, a NeuroMLDocument object or a Cell object

sanjayankur31:


sanjayankur31:
breaking since moving the vispy routines technically breaks API. I'm wondering if the helper methods should also be made private to prevent users from using these. They should only be using the high level `plot_interactive_3d` etc methods, not the lower level ones.

PR #242, opened 2023-07-17
Feat/chan den analysis

sanjayankur31:


sanjayankur31:
TODO: handle the `H` function. It needs to be replaced by sympy's Heaviside function, but by default, `H(0)=0.5` in sympy, so I've got to figure out how to replace `H(x)` with `Heaviside(x, 0)`.

PR #243, opened 2023-07-27
Feat/nsgr integration

sanjayankur31:


sanjayankur31:
For the moment, I used the HH example from docs to test it out. It was submitted and ran fine but threw up this error that I now need to debug:

```
  File "LEMS_HH_single_compartment_example_sim_nrn.py", line 202
    py_v_pop0_0_/v_output0 = [ float(x  / 1000.0) for x in h.v_pop0_0_/v_output0.to_python() ]  # Convert to Python list for speed, variable has dim: voltage
    ^
SyntaxError: cannot assign to operator
```

Can reproduce on my local machine too, will debug and see what's causing it.


TODO for this PR: add a test

sanjayankur31:
A simple test in the pyneuroml/examples directory:

```
python -c 'from pyneuroml.nsgr import run_on_nsg ; run_on_nsg("jneuroml_neuron", "LEMS_NML2_Ex5_DetCell.xml")'
```

fails because the output is expected in a `results` directory, but the simulation does run.

sanjayankur31:
TODO: tweak netpyne run

sanjayankur31:
This is now ready for review. It doesn't rely on newly generated files. It creates a new directory and does all its work there, so it should not be affected by any old files at all.

To test netpyne, in the `examples` folder, run (replace engine with `jneuroml_neuron` for neuron:

```
python -c 'from pyneuroml.nsgr import run_on_nsg ; run_on_nsg("jneuroml_netpyne", "LEMS_NML2_Ex5_DetCell.xml")'
```

Note that netpyne on nsgr is too old from the looks of it: https://github.com/NeuroML/NeuroML2/issues/213

PR #260, opened 2023-09-30
Improve performance of vispy plotter using meshes and their instances

sanjayankur31:
Includes some heuristics to decide which cylinders are considered instances of the same mesh---could be further tweaked.

### DIFF
--- a/pyneuroml/plot/PlotMorphology.py
+++ b/pyneuroml/plot/PlotMorphology.py
@@ -250,20 +250,26 @@ def plot_2D(
         print("Plotting %s" % nml_file)
 
     if type(nml_file) is str:
-        nml_model = read_neuroml2_file(
-            nml_file,
-            include_includes=False,
-            check_validity_pre_include=False,
-            verbose=False,
-            optimized=True,
-        )
-        # load bits we need to plot the model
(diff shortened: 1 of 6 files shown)


sanjayankur31:
> Is pyqt5 an actual requirement? 3D works fine on my mac now with that removed from requirements. Also requires vispy>=0.13.0

I think it is---vispy needs to have a "backend" and pyqt5/qt6 are probably the best ones:

https://vispy.org/installation.html#backend-requirements

Is it perhaps conflicting with another qt/pyqt install on your system---one installed using homebrew or something?

pgleeson:
I did have PyQt6 preinstalled. Any reason not to use that in the requirements?

sanjayankur31:
Not really. I know qt5 has been around for ages so is fairly stable. qt6 is newer, and it's available on Fedora now, but I don't know if it's as heavily used as qt5 yet. Could maybe do two extras: vispy-qt5 and vispy-qt6 to support both and let the user choose? I'd initially thought we'd let the user choose the backend and install it themselves so had not included pyqt5 at all, but then the first time they run our tool, they'll get nothing, so I added pyqt5 as the default.

sanjayankur31:
Updated deps to use pyqt6, and at least vispy 0.13.0.

PR #264, opened 2023-10-20
Close cylindrical meshes for vispy viewer

sanjayankur31:
Adapted from https://github.com/vispy/vispy/blob/main/vispy/geometry/generation.py#L451

I don't notice a performance hit, but it'll be good to test this out more just to be sure.


pgleeson:
Looks good!

sanjayankur31:
Cool, I'll cut a new release so this gets out to users too.

PR #287, opened 2023-12-06
added SEDML validate option

robertvi:
Also added a script to run the validation options on the SBML (with SEDML) test suite, but this was then moved into the SBMLShowcase repo, and so is no longer present here


robertvi:
note: I removed py3.7 from ci.yml, so this change probably duplicates the change @sanjayankur31 made

robertvi:
will do a new pull request for the rebased version

PR #294, opened 2024-01-05
feat: morphology plotting: allow marking/highlighting individual segments

sanjayankur31:
For ex: 

image

Useful to mark what segments are being used/recorded from.

Done in vispy, need to do in matplotlib viewer also.


sanjayankur31:
Also implemented for 2D (matplotlib)

image


With different marker size:

image

PR #295, opened 2024-01-31
added -sbml-sedml option passing through to jnml

robertvi:
A simple pass through of -sbml-sedml from pynml to jnml


sanjayankur31:
@robertvi : I've just merged into this to (hopefully) fix the CI. Please run a `git pull` in your local copy to get the latest version.

This looks fine to me, but I think we need to wait for a new jNeuroML release before merging this in, since that's required for the options to be available in jnml.

pgleeson:
@sanjayankur31 nope, -sbml-sedml has been an option in jnml for a long time...

sanjayankur31:
Cool, I'll cut a quick release then so all of this gets to users now.

PR #298, opened 2024-02-07
Adds util method to generate simulation scripts in a different, clean directory

sanjayankur31:
This is useful for various things like submitting to remotes, and for parameter sweeps etc. where one may be generating many variations of the model. Generating all the scripts in the same location by overwriting old files is error prone.


sanjayankur31:
We could maybe use a better function name than `generate_sim_scripts_in_folder`, but I couldn't think of anything off the top of my head that was descriptive enough.

sanjayankur31:
Looks fine, doesn't seem to break any APIs. All tests pass, and NMLlite looks happy too. Merging now and we'll fix any issues if they come up.

pgleeson:
Tested locally and all looks good!

Issue #278, opened 2023-11-29
Add option to save vispy generated meshdata 


sanjayankur31:
This will allow visualisation in other tools, such as [paraview](https://www.paraview.org/).

This should be simple enough, given that vispy provides the required I/O functions. We only have to save the mesh after we generate it

https://vispy.org/api/vispy.io.mesh.html#vispy.io.mesh.read_mesh

Dhruvanshu-Joshi:
Hi @sanjayankur31 ,
I would like to work on this issue. I did read the codebase and according to my understanding, the [PlotMorphologyVispy](https://github.com/NeuroML/pyNeuroML/blob/05801b3aeaf192883d5df8c5fc5f0f2533700724/pyneuroml/plot/PlotMorphologyVispy.py) currently uses vispy for visualisation. I believe that a solution to this would involve adding a new parameter to [plot_3D_cell_morphology](https://github.com/NeuroML/pyNeuroML/blob/05801b3aeaf192883d5df8c5fc5f0f2533700724/pyneuroml/plot/PlotMorphologyVispy.py#L590C1-L605C3). Based on the truth value of this parameter, we can make calls to new functions: save_meshdata_to_file which will use vispy's write_mesh function to save the mesh files. 
Is this what you had in mind and can you please assign this issue to me?

sanjayankur31:
Hi @Dhruvanshu-Joshi , yes that sounds correct. Do take a look at what common formats for storing mesh data are. I think the paraview docs do have some information on that. The ideal case would be that our code generates the mesh in a standard format so that any other tools can be used to then visualise it.

(A complementary tweak could also be to allow our methods to then read these meshes and visualise them)

Assigning this to you now. Thanks very much for working on this.

A few notes:

- do base your new branch against the development branch
- do use a virtualenv and run all the tests: run `pip install -e .[dev]` should install all the necessary deps, and `test.sh` should run them
- there's also pre-commit in use, so do run `pre-commit install` in your venv to get that set up
- do use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)

Dhruvanshu-Joshi:
Hi @sanjayankur31 ,
I am able to save the 3d mesh in ```.obj``` and ```.gz``` format only as these are the only formats supported by [vispy](https://github.com/vispy/vispy/blob/512a1ce22daccabc1a80356b33ab2d57cad86b22/vispy/io/mesh.py#L74) yet.

[...]

PR #316, opened 2024-03-11
<feat> : Added feature to save vispy obj and gz files

Dhruvanshu-Joshi:
**What is this PR about?**
This PR aims to solve issue [278](https://github.com/NeuroML/pyNeuroML/issues/278). Added parameter name "save_to_file" to "plot_interactive_3d". Also added a test.

**Checklist**
+ [x] Explain important implementation details 👆
+ [x] Make sure that [the pre-commit linting/style checks pass]
+ [x] Link relevant issues 
+ [x] Are the changes covered by tests and docstrings?


sanjayankur31:
Thanks, I'll review it in the next couple of days

sanjayankur31:
@Dhruvanshu-Joshi ^ to begin with, would you please investigate why the tests are failing? (no hurry, please take your time, your coursework/exams are more important)

Dhruvanshu-Joshi:
Thank you so much @sanjayankur31 . I'll look into this asap.

sanjayankur31:
How is this going @Dhruvanshu-Joshi ? Anything I can help with?

@all-contributors please add @Dhruvanshu-Joshi for bug, ideas

Dhruvanshu-Joshi:
> How is this going @Dhruvanshu-Joshi ? Anything I can help with?
> 
> @all-contributors please add @Dhruvanshu-Joshi for bug, ideas

Hi @sanjayankur31 , the error seems to be something with github actions.
Screenshot from 2024-03-20 15-47-47
Here it states that TEMPLATE.channel.nml does not exist. However, I checked both locally and on the main repo on github, the file does exist. Maybe it is not rendering on github actions. Can you help me with this? Also I did run the test locally and it seems to work fine.

sanjayankur31:
Ah, is your branch based on the `development` branch or `master`? It needs to be against `development`. The PR here also should be using `development` as the base branch---try updating that and maybe that'll solve the issue.

Dhruvanshu-Joshi:
Extremely sorry for this mistake. I'll rectify this and push the changes soon.

sanjayankur31:
Thanks, I've just returned from leave, so I'll take a look this week.

pyneuroml/utils/plot.py
```diff
@@ -389,3 +390,184 @@ def load_minimal_morphplottable__model(
...
+
+    # Create directories if they don't exist
+    os.makedirs(os.path.dirname(filename), exist_ok=True)
+
+    vertices = []
+    faces = []
+    normals = []
+    texcoords = []
+
+    for key, segments in meshdata.items():
```
sanjayankur31, line 438: So, we already create meshes in `create_instanced_meshes`, line 988, but we don't track them. Can we not just store these in a variable and then get the vertices etc. from this Meshdata instead of you basically recreating the mesh here which will be quite expensive for larger cells? 

See: https://vispy.org/api/vispy.geometry.meshdata.html#module-vispy.geometry.meshdata

From the test, it also appears that all the arguments are not necessary to generate the mesh:

https://github.com/vispy/vispy/blob/5d61ecbd930167446be0f9e5e078288e19839af9/vispy/io/tests/test_io.py#L67

Did you generate a mesh and see that it works in other tools? (Or did you just read it back in using vispy to see if it works?)

pyneuroml/utils/plot.py
```diff
@@ -389,3 +390,184 @@ def load_minimal_morphplottable__model(
                 for acell in inc.cells:
                     acell.biophysical_properties = None
                     nml_model.add(acell)
+
+
+def extract_coordinates_and_diameter(point_with_diam):
+    """
+    Extracts coordinates and diameter from a Point3DWithDiam object.
```
sanjayankur31, line 397: For all doc strings, please follow the format already being used in the code. We just use the default reST format (not Google or NumpyDoc). See: https://www.sphinx-doc.org/en/master/usage/domains/python.html#info-field-lists

tests/plot/test_morphology_plot.py
```diff
@@ -243,7 +243,12 @@ def test_3d_schematic_plotter(self):
     def test_3d_morphology_plotter_vispy_network(self):
         """Test plot_3D_cell_morphology_vispy function."""
         nml_file = "tests/plot/L23-example/TestNetwork.net.nml"
-        plot_interactive_3D(nml_file, min_width=1, nogui=True, theme="dark")
+        ofile = pl.Path(nml_file).name
+        plane = "xyz"
+        filename = f"tests/plot/test_morphology_plot_3d_{ofile.replace('.', '_', 100)}_{plane}_schematic.obj"
+        plot_interactive_3D(nml_file, min_width=1, save_to_file=filename, theme="dark")
+        self.assertIsFile(filename)
```
sanjayankur31, line 250: I'd create a new test function specifically for storing the mesh data. If possible, like in the vispy docs, we also want to open the file to see that it's contents are somewhat correct (not just that the file exists).

sanjayankur31 (changes requested):
Getting there, but we can probably simplify it by using the already generated mesh data. Take a look at the review comments.

sanjayankur31:
Hi @Dhruvanshu-Joshi , how are you doing? Any updates/progress on this? (If you don't have time to work on this, please let us know and we'll merge your changes and work from there)

Dhruvanshu-Joshi:
Hi @sanjayankur31 
Extremely sorry that I have delayed this so much. I am currently in midst of my exams but I'll like to try this again in June. Will it be fine?

sanjayankur31:
Sure, that sounds good. All the best for your exams!

pyneuroml/utils/plot.py
Dhruvanshu-Joshi: Are you suggesting to use a ```mesh object``` instead of ```meshdata``` to store the mesh?
I did not explicitly generate a custom mesh to test this. I read a 3d mesh using vispy and tested it.

sanjayankur31: So, if you look at our plotting code, we create cylindrical and spherical meshes ourselves. I am wondering if we can't just use that rather than you re-creating meshes. All of these meshes are stored as Meshdata---see `create_cylindrical_mesh` and the `meshdata` variable, which is passed to the different functions and holds all the meshdata we create

sanjayankur31:
Refactored and merged with our new implementation here: https://github.com/NeuroML/pyNeuroML/commit/fe9b3ab5fb1556ff819a99a9a4fd2a9db8b0d3e4

PR #326, opened 2024-03-24
add -swc in pynml

### LINKED ISSUE #133: Add `-swc` option to export NeuroML cell files to SWC format
This is already implemented in the `pyneuroml.swc` package, so it just just needs to be exposed.

YBCS:
closes #133 
- adds a new cli arg in pynml to allow converting `NeuroML` to swc `format`


YBCS:
@sanjayankur31 PTAL

sanjayankur31:
Looks good @YBCS  . Let's add some tests etc. to make sure this is always checked:

- can you please add a test command in `test.sh` where it says "test some conversions"? You can use the example files in `examples`, the ones that have been used in `ExportSWC.py`
- it seems to print every line during the conversion, can you turn that off---for large conversions, that'll be a lot of output. Could you convert the print statement to a `logger.debug` perhaps?
- could you look at generating the man pages again? you need to run the `generate-man-pages.sh` script, you'll need to have `help2man` installed
- could you check the doc-string for the `-swc` option you've added? Can one provide multiple files? If yes, that doc string needs to reflect that.

sanjayankur31 (changes requested):
Requesting changes

YBCS:
@sanjayankur31 I have addressed the comments please take a look !

YBCS:
@sanjayankur31 who should I ask for review ?

sanjayankur31:
I've just returned from leave, so I'll take a look this week.

pyneuroml/swc/ExportSWC.py
```diff
@@ -124,6 +124,9 @@ def convert_to_swc(nml_file_name, add_comments=False, target_dir=None):
     lines = []
     comment_lines = []
 
+    if not nml_doc.cells:
```
sanjayankur31, line 127: From what I remember, `nml_doc` will always have `cells`, but it can be an empty list---so perhaps we need to check the length of this here instead.

sanjayankur31 (changes requested):

pyneuroml/swc/ExportSWC.py
YBCS: okay, I updated the check to check len of `nml_doc.cells` instead

common.h2m
```diff
@@ -0,0 +1,16 @@
+[see-also]
```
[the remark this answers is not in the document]
YBCS: I deleted the files in top level directory and ran the script in `/man/man1`, however all files are regenerated anyway because the script tries to generate man file for all files and not just a specific file

sanjayankur31:
I think that looks good now. There are a few issues with the exporter, but I'll file fresh issues for them separately. Let's wait for CI to finish, and then I can merge this.

A few notes: we can write better git commit messages. For example "address comments" doesn't mean anything---in the future if you come across that commit, it won't tell you anything. Take a look at this post on writing good git commit messages:

https://cbea.ms/git-commit/

and this one on conventional commits---we don't enforce conventional commits, but they would be good to use since they really improve the readability of commits:

https://www.conventionalcommits.org/en/v1.0.0/

sanjayankur31 (approved):

sanjayankur31:
@all-contributors please add @YBCS for code

YBCS:
> I think that looks good now. There are a few issues with the exporter, but I'll file fresh issues for them separately. Let's wait for CI to finish, and then I can merge this.
> 
> A few notes: we can write better git commit messages. For example "address comments" doesn't mean anything---in the future if you come across that commit, it won't tell you anything. Take a look at this post on writing good git commit messages:
> 
> https://cbea.ms/git-commit/
> 
> and this one on conventional commits---we don't enforce conventional commits, but they would be good to use since they really improve the readability of commits:
> 
> https://www.conventionalcommits.org/en/v1.0.0/

@sanjayankur31 Thanks, I will make the commit message more clear next time.

PR #332, opened 2024-04-11
feat(LEMSSimulation): allow using float/str for duration/dt

### LINKED ISSUE #199: Update LEMSSimulation to also take string parameters
`duration` and `dt` currently have to be provided as floats, which is not in line with the rest of neuroml where all parameters are provided as strings with units. So we should enhance the LEMSSImulation class to also allow inputting these values as strings with units for consistency.

sanjayankur31:
It's more in line with the rest of NeuroML where we provide values as strings with units.

Fixes #199


pgleeson:
Looks good to me!

PR #334, opened 2024-04-13
Issue/318- Adding Video recording functionality to vispy viewer

AdityaBITMESRA:
I have modified the code in the create_new_vispy_canvas function to add video recording functionality

### DIFF
--- a/pyneuroml/plot/PlotMorphologyVispy.py
+++ b/pyneuroml/plot/PlotMorphologyVispy.py
@@ -14,6 +14,11 @@ import logging
 import math
 import random
 import typing
+import imageio
+from vispy.gloo import util
+import vispy.app
+from vispy.scene import SceneCanvas
+from vispy.visuals import MeshVisual
 
 import numpy
 import progressbar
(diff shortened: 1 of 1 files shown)


pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -120,6 +125,14 @@ def create_new_vispy_canvas(
...
+    canvas_name = "My 3D Visualization"
+    canvas, scene_canvas, view = create_new_vispy_canvas(canvas_name)
+    canvas = vispy.app.Canvas(keys="interactive", size=(800, 600), title=canvas_name)
+    scene_canvas = SceneCanvas(keys="interactive", show=True)
+    view = scene_canvas.central_widget.add_view()
+
+    camera = scene_canvas.central_widget.camera
+    view.camera = camera
 
     canvas = scene.SceneCanvas(
```
sanjayankur31, line 137: we're already creating a SceneCanvas, camera, view etc. here---do you need to create one too? Can the ones we already create not be used?

sanjayankur31 (changes requested):
Looking good, but needs a few changes. My main question is---can the canvas etc. we already create not be used? Why is a new one required?

Also: in the future, I wouldn't use the `development` branch for PRs. It's best to create a new branch from `development` and use that. That way your `development` branch always remains clean and matches our upstream `development` branch.

pyneuroml/plot/PlotMorphologyVispy.py
AdityaBITMESRA: yes sir you are correct I think I ignored this.

Issue #279, opened 2023-11-29
Include generation of animated matplotlib plots


sanjayankur31:
Matplotlib allows animations, this could be useful when people are looking at the temporal evolution of recorded data:

https://matplotlib.org/stable/users/explain/animations/animations.html#sphx-glr-users-explain-animations-animations-py

sanjayankur31:
https://spikesandbursts.wordpress.com/2024/01/04/patch-clamp-data-analysis-animate-time-series/

https://matplotlib.org/stable/api/_as_gen/matplotlib.animation.FuncAnimation.html

YBCS:
@sanjayankur31 can I work on this ?

sanjayankur31:
Sure, assigning to you now.

YBCS:
@sanjayankur31 there are many plotting functions, should  I start with one in particular or should I try to implement something generic and can work with many different types of data.

sanjayankur31:
Ideally, I think we'd like to add an argument to the `generate_plot` to make it also generate animated matplotlib plots. So, maybe start there?

YBCS:
@sanjayankur31 
[Screencast from 24-04-24 12:49:46 AM IST.webm](https://github.com/NeuroML/pyNeuroML/assets/29734619/9bcc2443-8600-4937-89c8-486d46980f01)
I got the animation to work but there are some limitations
I wanted to confirm before I raise my PR 

so this will only work if `show_plot_already` is `True`. The reason is that for animating we have to write an animate function and it needs to be passed to `FuncAnimation` and this needs to happen in the same context in which plt.show() is called otherwise the canvas is wiped and nothing gets plotted.

sanjayankur31:
OK, I think that's good for the first iteration. Later, we can see if there's a way of recording the animation (maybe as a gif/mp4), with and if possible in matplotlib, without showing the window.

What we could do is add a new "animate" argument to the function, and if that is given, we toggle "show_plots_already" to true? How does that sound?

As an additional note, in another PR, there's also code to make a video from multiple "frames". Maybe we could use that in other plotting code (like a wrapper around `plot_2d_cell_morphology`)  to generate multiple images and then stitch them together into a video?

https://github.com/NeuroML/pyNeuroML/pull/342/files

YBCS:
> OK, I think that's good for the first iteration. Later, we can see if there's a way of recording the animation (maybe as a gif/mp4), with and if possible in matplotlib, without showing the window.
> 
[...]

PR #348, opened 2024-04-23
animate matplotlib plots

YBCS:
#279 wip fix 

we can only use this feature if `show_plot_already` is set to `True`


sanjayankur31 (changes requested):
Looking good. I've added a test and opened a PR for you to add that to your branch: 

https://github.com/YBCS/pyNeuroML/pull/1

Can we also add saving of animations? Looks simple enough? 

https://matplotlib.org/stable/users/explain/animations/animations.html#funcanimation

pyneuroml/plot/Plot.py
```diff
@@ -240,6 +246,21 @@ def generate_plot(
         logger.info("Saved image to %s of plot: %s" % (save_figure_to, title))
 
     if show_plot_already:
+        if animate:
+            duration = 5
```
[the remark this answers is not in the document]
YBCS: duration varibale is not consistent. "duration" will depend of lots of factors including data calculation and actual length of data to plot and machine configuration. Even If I try to calculate the "hyperparameter" variables to make it perform the animation in required duration, the actuall outcome usually differs.

pyneuroml/plot/Plot.py
```diff
@@ -240,6 +246,21 @@ def generate_plot(
...
+            num_frames = len(xvalues[0])
+            interval = duration * 1000 / num_frames
```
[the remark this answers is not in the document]
YBCS: yes, I will update this. 
Interval is simply " Delay between frames in milliseconds", it can be harcoded. Most tutorial I have seen uses 50 or 100 ms.

YBCS:
> Can we also add saving of animations? Looks simple enough?
> 
> https://matplotlib.org/stable/users/explain/animations/animations.html#funcanimation

Saving animation is simple. But after some testing with the plots in the codebase I have identified some issues. 

Animation is actually very expensive and so it takes lots of time. I have encountered plots with >= 10000 ([link](https://github.com/NeuroML/pyNeuroML/blob/3269c90d1cf8f22fad9c3225945cb5159911f067/examples/quick_plot.py#L12)) 
In my machine saving animation plot with data length 1000 can take more almost a minute. 
Plot is a generic function, it is probably not a good idea if the plot will have lots of data. 

On the other hand the script can choose not to save animation if the provided data point is too long.
1000 is a good threshold as it can save the animation within 1 minute.

tests/plot/test_plot.py
```diff
@@ -19,9 +21,39 @@
...
+        except FileNotFoundError:
+            pass
+
+        numpoints = 100
+        xs = list(range(0, numpoints))
+        ys = random.choices(list(range(0, 1000)), k=numpoints)
+        ys2 = random.choices(list(range(0, 1500)), k=numpoints)
+
+        generate_plot(
+            [xs, xs],
```
YBCS, line 43: [xs1, xs2] is it possible that this list contains items with differing lengths. can xs1 and xs2 have different lengths ?

pyneuroml/plot/Plot.py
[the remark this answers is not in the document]
sanjayankur31: Cool--yeh, I think 50ms (20fps) is good for a default minimum value.

tests/plot/test_plot.py
sanjayankur31: Yes, I think they can, but [y1, y2] must have the same lengths as [x1, x2] so that matplotlib has complete (x,y) co-ordinates for each point. 

For animation, one will have to find the values with the longest length and use that to calculate duration and all that then?

pyneuroml/plot/Plot.py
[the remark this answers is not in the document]
sanjayankur31: So, if we say duration "5", that doesn't mean that the animation will be 5s long? :thinking:

pyneuroml/plot/Plot.py
```diff
@@ -240,6 +249,41 @@ def generate_plot(
...
+                cache_frame_data=False
+            )
+
+            frame_length_threshold = 5000
+            if len(xvalues[0]) < frame_length_threshold and save_animation_to:
+                logger.info("Saving animation to %s" %
+                            (save_animation_to))
+                ani.save(
+                    filename=save_animation_to,
+                    writer="pillow",
```
sanjayankur31, line 283: I wonder if we should make this an argument too, to allow the user to choose different writers? ffmpeg/imagemagick may be quicker than pillow and support more formats:

https://matplotlib.org/stable/users/explain/animations/animations.html

How about making it an argument of the form:

```
writer={ 'writer name': ["extra args list"] }
```

with default value:

```
writer = { "pillow": [] }
```

?

sanjayankur31: We should also add this link to the docstring around these arguments:

https://matplotlib.org/stable/users/explain/animations/animations.html#saving-animations

pyneuroml/plot/Plot.py
```diff
@@ -240,6 +249,41 @@ def generate_plot(
...
+                fig=fig,
+                frames=size-1,
+                func=update,
+                interval=interval,
+                blit=True,
+                cache_frame_data=False
+            )
+
+            frame_length_threshold = 5000
+            if len(xvalues[0]) < frame_length_threshold and save_animation_to:
```
sanjayankur31, line 278: This is to limit animations to 5000 frames only? If this depends on the users' setups (how good their hardware etc. is), let's not add this check like this. Maybe we can just add a warning that is thrown if there are more than 5000 points that says "too many data points, this may take a while"? That way the user can still go ahead and use it if they wish. Maybe they'll start it and go eat lunch, and it'll be complete by the time they return (we do this a lot with our simulations :))

pyneuroml/plot/Plot.py
```diff
@@ -240,6 +249,41 @@ def generate_plot(
         logger.info("Saved image to %s of plot: %s" % (save_figure_to, title))
 
     if show_plot_already:
+        if animate:
+            d = 5
```
sanjayankur31, line 253: I need to test this out more. I thought the idea was that if I set duration to 5000ms, the animation would be 5000ms, but I think you're saying that this is not the case?

sanjayankur31 (commented):
Almost there, I just have a few things I'm unclear on.

pyneuroml/plot/Plot.py
YBCS: yes, that is what is happening with my current implementation. 
My calculations could be improved I think and it should be possible to atleast get as close as possible to set duration, let me try again.

pyneuroml/plot/Plot.py
YBCS: I see. I will add a warning and it will ask for user input to continue with the saving or not (possibly showing approx time (minutes) it might take to complete the entire process. 
It will then save or discard based on the user input. 
How does that sound ?

pyneuroml/plot/Plot.py
YBCS: should I include this in the same PR or should it be done in another PR.

tests/plot/test_plot.py
YBCS: >> "For animation, one will have to find the values with the longest length and use that to calculate duration and all that then?"

Yeah that will be correct as, [plt1, plt2] are individual plots and we only care of the total duration which means the max of the lengths

pyneuroml/plot/Plot.py
sanjayankur31: I'd skip the user input entirely. Just make it an argument and if it's more than something, just print a warning saying "this could take a while". If it takes too long, the user can always interrupt it and use a smaller value.

(Adding interaction to take user input will make it trickier to do batch analyses, so we want to avoid that as much as possible)

pyneuroml/plot/Plot.py
sanjayankur31: I think we can roll this in here too?

pyneuroml/plot/Plot.py
sanjayankur31: Don't worry about getting it exactly right---close enough will do and it can be tweaked later

sanjayankur31 (approved):
I think that looks good. I'll just test it locally once and then merge :sparkles:

sanjayankur31:
LGTM, merging. I'll make a few tweaks in my branch and merge to development :+1:

PR #352, opened 2024-04-26
feat(pynml-modchananalysis): add labels and fix file name

sanjayankur31:
To ensure that the graphs from modchananalysis and pynml-chananalysis are identical.


sanjayankur31:
@pgleeson : there wasn't a reason why we'd left the axes labels out from the plots, was there?

pgleeson:
LGTM!

PR #361, opened 2024-05-16
feat(rotation): allow users to pick if they want a copy

sanjayankur31:
Adds an option to allow users to rotate cell without creating a copy. 

@lej0hn : can do this in the translate function too, so that if the two operations are run one after the other, we don't need to create a copy each time.


lej0hn (approved):
Will do

sanjayankur31:
@all-contributors please add @lej0hn for review

Issue #360, opened 2024-05-16
Add utility function to translate cell to that soma is at origin


sanjayankur31:
Is of general use

PR #367, opened 2024-05-18
Added translate_cell function

lej0hn:
Aims to solve #360 . Function translates cell so that its soma at the origin of the axes

### DIFF
--- a/pyneuroml/utils/__init__.py
+++ b/pyneuroml/utils/__init__.py
@@ -499,9 +499,9 @@ def translate_cell_to_coords(
         [soma_seg.proximal.x, soma_seg.proximal.y, soma_seg.proximal.z]
     )
 
-    translation_x = cell_origin[0] - x
-    translation_y = cell_origin[1] - y
-    translation_z = cell_origin[2] - z
+    translation_x = cell_origin[0] - dest[0]
+    translation_y = cell_origin[1] - dest[1]
+    translation_z = cell_origin[2] - dest[2]
 
     if translation_x == translation_y == translation_z == 0:
         return cell
@@ -512,7 +512,7 @@ def translate_cell_to_coords(
         newcell = cell
 
 
-    logger.info(f"Translating {newcell.id} by x:{translation_x}, y:{translation_y}, z:{translation_z}")
+    logger.info(f"Translating {newcell.id} by x:{-translation_x}, y:{-translation_y}, z:{-translation_z}")
     
     # translate each segment
     for aseg in newcell.morphology.segments:
--- a/pyneuroml/utils/__init__.py
+++ b/pyneuroml/utils/__init__.py
@@ -476,25 +476,19 @@ def rotate_cell(
 def translate_cell_to_coords(
     cell: neuroml.Cell,
     inplace: bool = False,
-    x: float = 0,
-    y: float = 0,
-    z: float = 0,
+    dest : typing.List[float] = [0,0,0],
 ) -> neuroml.Cell:
     """Translate cell so that its soma moves to given coordinates
 
     .. versionadded:: 1.2.13
-    
+
     :param cell: cell object to translate
     :type cell: neuroml.Cell
     :param inplace: toggle whether the cell object should be modified inplace
         or a copy created (creates and returns a copy by default)
     :type inplace: bool
-    :param x: x axis point destination of soma
-    :type x: float
-    :param y: y axis point destination of soma
-    :type y: float
-    :param z: z axis point destination of soma
-    :type z: float
+    :param dest: destination coordinates (x,y,z) for cell's soma
+    :type dest: list[x,y,z]
     :returns: new neuroml.Cell object
     :rtype: neuroml.Cell
 
--- a/pyneuroml/utils/__init__.py
+++ b/pyneuroml/utils/__init__.py
@@ -473,6 +473,61 @@ def rotate_cell(
 
     return newcell
 
+def translate_cell(
+    cell: neuroml.Cell,
+    inplace: bool = False,
+) -> neuroml.Cell:
+    """Return a new cell object translated so that the some is at the origin.
+
+    :param cell: cell object to rotate
+    :type cell: neuroml.Cell
+    :param inplace: toggle whether the cell object should be modified inplace
+        or a copy created (creates and returns a copy by default)
+    :type inplace: bool
+    :returns: new neuroml.Cell object
+    :rtype: neuroml.Cell
+
+    """
+    soma_seg_id = cell.get_morphology_root()
+    soma_seg = cell.get_segment(soma_seg_id)
+    cell_origin = numpy.array(
+        [soma_seg.proximal.x, soma_seg.proximal.y, soma_seg.proximal.z]
+    )
+
+    if cell_origin[0] == cell_origin[1] == cell_origin[2] == 0:
+        return cell
+    
+    if not inplace:
+        newcell = copy.deepcopy(cell)
+    else:
+        newcell = cell
+
+
+    print(f"Translating {newcell.id} by x:{cell_origin[0]}, y:{cell_origin[1]}, z:{cell_origin[2]}")
+    
+    # translate each segment
+    for aseg in newcell.morphology.segments:
+        prox = dist = numpy.array([])
+        # may not have a proximal
+        try:
+            prox = numpy.array([aseg.proximal.x, aseg.proximal.y, aseg.proximal.z])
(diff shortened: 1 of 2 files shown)


pgleeson:
Perhaps a better name for this would be `translate_cell_to_origin()` as there could be another method `translate_cell()` which takes a point as an argument and translates it to there..?

pyneuroml/utils/__init__.py
```diff
@@ -473,6 +473,61 @@ def rotate_cell(
 
     return newcell
 
+def translate_cell_to_origin(
+    cell: neuroml.Cell,
+    inplace: bool = False,
+) -> neuroml.Cell:
+    """Return a new cell object translated so that the some is at the origin.
```
sanjayankur31, line 480: "Translate cell to origin" is probably a better description. the `inplace` param tells users that they can use the same object if they wish. What do you think?

pyneuroml/utils/__init__.py
```diff
@@ -473,6 +473,61 @@ def rotate_cell(
...
+    if cell_origin[0] == cell_origin[1] == cell_origin[2] == 0:
+        return cell
+    
+    if not inplace:
+        newcell = copy.deepcopy(cell)
+    else:
+        newcell = cell
+
+
+    print(f"Translating {newcell.id} by x:{cell_origin[0]}, y:{cell_origin[1]}, z:{cell_origin[2]}")
```
sanjayankur31, line 506: Should this be a debugging comment? `logger.debug`?

pyneuroml/utils/__init__.py
```diff
@@ -473,6 +473,61 @@ def rotate_cell(
...
+    
+    # translate each segment
+    for aseg in newcell.morphology.segments:
+        prox = dist = numpy.array([])
```
sanjayankur31, line 510: is `dist` being used somewhere?

lej0hn:
Yes of course, fixed it

pyneuroml/utils/__init__.py
```diff
@@ -473,6 +473,61 @@ def rotate_cell(
 
     return newcell
 
+def translate_cell_to_origin(
+    cell: neuroml.Cell,
+    inplace: bool = False,
+) -> neuroml.Cell:
+    """Return a new cell object translated so that the some is at the origin.
+
+    :param cell: cell object to translate
```
sanjayankur31, line 482: Let's add an `versionadded` info bit here too, so users know what version this is available from.

I think it'll be:

``` rst
.. versionadded:: 1.2.13

```

(search the sources for `versionadded` just to be sure)

sanjayankur31 (changes requested):
Looks good. A few cosmetic comments.

Let's also add a test case? You can construct a small cell with a few segments and then check that they have the right info after they've been translated? (See the `test_rotate_cell` method in the `test_utils.py` file.)

pyneuroml/utils/__init__.py
lej0hn: That's true, we should probably change that at the rotate_cell() function as well

pyneuroml/utils/__init__.py
lej0hn: Perhaps, got it from rotate_cell() function and though it might be useful

pyneuroml/utils/__init__.py
lej0hn: Hm, i'm not familiar with this, could you please elaborate a bit what I need to look for?

lej0hn:
Yes, I think that's a good idea

pyneuroml/utils/__init__.py
sanjayankur31: Take a look at this: https://github.com/NeuroML/pyNeuroML/blob/bfb0197943efbff74e6dc1fee3cd102e6d2433d7/pyneuroml/runners.py#L1264

and the generated doc:

https://pyneuroml.readthedocs.io/en/development/pyneuroml.pynml.html#pyneuroml.pynml.generate_sim_scripts_in_folder

Sphinx docs: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded

lej0hn: Ok, got it, so since currently we are on v1.2.12, I should the next one which will be v1.2.13, right?

sanjayankur31: Yes. We bump the release in setup.cfg to the next version after making a release. So that's at 1.2.13 now.

lej0hn:
This is ready for review. I changed the name to translate_cell_to_coords thinking there might be a future function, translate_by for example, that could translate the cell by some distance (x,y,z) in the according axis

tests/utils/test_utils.py
```diff
@@ -156,3 +157,92 @@ def test_rotate_cell(self):
 
         newdoc.validate(recursive=True)
         write_neuroml2_file(newdoc, "tests/utils/test_rotation.net.nml", validate=True)
+
+    def test_translate_cell_to_coords(self):
+            """Test translate_cell_to_coords"""
+            acell = neuroml.utils.component_factory(
+                "Cell", id="test_cell", validate=False
+            )  # type: neuroml.Cell
+
```
sanjayankur31, line 166: add a spike threshold here, with the latest release of the schema (currently in the dev branches), spike thresh will become a compulsory element:

```
acell.set_spike_thresh("10mV")
```

tests/utils/test_utils.py
```diff
@@ -156,3 +157,92 @@ def test_rotate_cell(self):
...
+                seg_id=3,
+                use_convention=False,
+                reorder_segment_groups=False,
+                optimise_segment_groups=False,
+                parent=soma,
+            )
+
+            print(f'cell before: {acell}')
+
+            translated_cell = translate_cell_to_coords(acell, False, [30,30,30])
```
sanjayankur31, line 208: how about an assertion to test that the soma is now at `[30, 30, 30]`, and that a few of the segments have moved by the correct amount? (I couldn't figure out an easy way to test out rotation, but we should be able to have a few basic checks for translation?)

sanjayankur31 (changes requested):
looks very good, a few minor tweaks/comments to look at

pyneuroml/utils/__init__.py
```diff
@@ -494,7 +499,11 @@ def translate_cell_to_origin(
         [soma_seg.proximal.x, soma_seg.proximal.y, soma_seg.proximal.z]
     )
 
-    if cell_origin[0] == cell_origin[1] == cell_origin[2] == 0:
+    translation_x = cell_origin[0] - dest[0]
```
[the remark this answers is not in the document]
lej0hn: Don't really have a reason, is there a difference?

tests/utils/test_utils.py
lej0hn: Should I also add it to the test_rotate_cell function?

lej0hn: After adding that and running the test I get some warnings like the following, is it normal?
tests/utils/test_utils.py::TestUtils::test_translate_cell_to_coords                                                       
:..site-packages\neuroml\nml\generatedssupersuper.py:176: UserWarning: membrane_properties has already been assigned.  Use `force=True` to overwrite. Hint: you can make changes to the already added object as required without needing to re-add it because only references to the objects are added, not their values.

pyneuroml/utils/__init__.py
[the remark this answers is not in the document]
sanjayankur31: `dest - src` is just more intuitive I guess, but it doesn't matter here as long as its clear.

tests/utils/test_utils.py
sanjayankur31: Yes, that's fine. It's a message we give to the user to tell them that a part of the model already exists and we didn't overwrite it

sanjayankur31 (approved):
LGTM, merging

sanjayankur31:
@all-contributors please add @lej0hn for code

sanjayankur31:
@lej0hn : quick question: do you have pre-commit setup so that the files are formatted before being committed? Do check because my pre-commit formatted both files again when I looked at them, suggesting they weren't originally formatted here in the PR

lej0hn:
No, I don't have pre-commit setup. I will check that

sanjayankur31:
Cool, all you have to do is run `pre-commit install` in the repo, and it'll take care of things automatically after that.

https://github.com/NeuroML/pyNeuroML/blob/development/CONTRIBUTING.md#pre-commit

PR #379, opened 2024-06-03
Feature Draft/ Use PCA to visualize cell  'upwards'

lej0hn:
- Finds the first PCA
- Finds rotations needed to bring it along z axis
- Rotates the cell using same angles
- Translates cell to origin
- Visualizes axes and PCA axis (before and after rotation)


sanjayankur31:
OK for me to take a peek?

lej0hn:
Yes, go ahead!

lej0hn:
The code is in different functions right now and I added the extra parameters just to be able to temporarily visualize the pca axis as well

sanjayankur31:
Cool, on my list, will hopefully have a look today/tomorrow :+1:

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -471,6 +486,50 @@ def plot_interactive_3D(
 
     else:
         cell = list(pop_id_vs_cell.values())[0]
+        #Get all segments' distal points
+        segment_points = []
```
sanjayankur31, line 490: I wonder if we can move this functionality into the `plot_3D_cell_morphology` function, and add an argument there to activate this transformation (instead of passing `x_angle` and `y_angle` pass: `pcaify=True` (with a better argument name, of course :))? I wonder because this is limited to the case where we're plotting a single cell, so it sort of fits better there. One should also be able to use this when using `plot_3D_cell_morphology` directly?

It'll also mean that the rotation + translation only happens once. At the moment, in the hay model, for example, it's happening multiple times---each time the mesh precision is dropped.

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -769,6 +832,9 @@ def plot_3D_cell_morphology(
     except Exception:
         axon_segs = []
 
+    cell = rotate_cell(cell, x_angle, y_angle, 0, 'yxz', False, False)
+    cell = translate_cell_to_coords(cell, False, [0,0,0])
```
sanjayankur31, line 836: I assume the order of operations is important---would it be different if we:

- first translated the cell to origin
- did the pca
- rotated the cell about the soma (origin)?

sanjayankur31 (commented):
Looking very good! I think we're almost there. I've dropped a few minor comments.

I think the camera isn't still pointing to origin, and we're still slightly off---see the rotation bit:

[Screencast from 2024-06-04 12-25-47.webm](https://github.com/NeuroML/pyNeuroML/assets/102575/ab47d6ae-a523-45ec-9c42-b5632c55b14e)

I'm not sure why this is given that youre axes are shown about the origin. Maybe the bounding box calculation is off by a bit, so the camera is slightly askew too?

pyneuroml/plot/PlotMorphologyVispy.py
lej0hn: Yes this is the goal, this just allowed for visualization of the pca axes for debugging. However, besides not being able to visualize the pca axes after this change, I think there is another issue with the camera, as you noted above. So, the bounding box that's used to center the camera is taken in `plot _interactive_3d` , before the cell is translated to origin or rotated, so the bounding box and therefore the camera, is centered on the original cell's position. Since in this case we're translating the cell to the origin, maybe along with `pcaify=True` , we could set an extra argument in `create_new_vispy_canvas` that sets the `view_center`  of the camera at the origin as well?

pyneuroml/plot/PlotMorphologyVispy.py
lej0hn: - first translated the cell to origin
- did the pca
These two steps can be done in any order ,I believe, without any change because the PCA first translates the points so that their center is at the origin and then applies other computations.

But after that, the best way is in fact to translate and then rotate, now that I think of it, because the rotations and the angles are calculated based on the PCA axis, which starts at the origin. So to apply them succesfully  to the cell, it should be close to the origin in order to have similar results.

pyneuroml/plot/PlotMorphologyVispy.py
sanjayankur31: Yeh. So, if there's only one cell, we don't need to call `create_new_vispy_canvas` in `plot_3D_interactive` at all. I can't remember why I did it there. It doesn't seem necessary. We could conditionalise it there so that it's only called when there's more than one cell to be plotted, otherwise we let `plot_3D_cell_morphology` or `plot_3D_schematic` do it when it finds that `current_canvas` and `current_view` are `None`.

I just remembered that we also have the `plot_3D_schematic` function too, where it'll also be good to apply the PCA method so that the schematic is also upright. So perhaps we need to split out the PCA bits into a new method that takes a cell as argument and returns the "upright" cell? This could be called by both `plot_3D_cell_morphology` and `plot_3D_schematic` before they create the canvas and the meshes etc?

Yes, we could add another optional argument to `create_new_vispy_canvas` called `view_center`, which if provided will be used. If it isn't, we'll calculate the centre from `view_min` and `view_max`?

How does that sound?

lej0hn:
I'm facing a problem with creating a new canvas in functions `plot_3d_morphology` and `plot_3d_schematic` . It seems that due to the recursion which occurs to reduce precision, the canvas created in those functions (which happens in case the canvas from `plot_3d_interactive` is none) , gets overwritten and can't be used

sanjayankur31:
Hrm---I think if one uses the `plot_3D_morphology` etc. functions directly instead of `plot_intereactive_3D`, we don't check for precision and reduce it etc. So we don't recurse :thinking: 

I'll try to take a look at your PR today, otherwise we can discuss it at the meeting tomorrow and see what to do.

lej0hn:
I didn't check for using them directly, so you are probably right at that. Perhaps it is a bit of a hustle, however, for the user, to call these functions directly instead of handling it through `plot_interactive_3d`?

sanjayankur31:
- test out with cells: https://github.com/OpenSourceBrain/AllenInstituteNeuroML/tree/master/CellTypesDatabase/models
- align according to y axis
- tweak camera to make Y up in general

sanjayankur31:
- give unique colours to axes for both themes

lej0hn:
I am facing an issue with the rotation we discussed around the y axis. There was no problem getting the camera to look at the y axis and make the cell align to it, but I can not find a way to make the camera orbit around the y axis. The orbit function has only 2 arguments, elevation and azimuth which rotate it around the x and z axis respectively ( https://vispy.org/api/vispy.scene.cameras.turntable.html#vispy.scene.cameras.turntable.TurntableCamera.orbit). Also the XYZ shows the axis but doesn't use different colors. Any ideas?

sanjayankur31:
> I used XYZ from vispy and made it work, but the colors are all the same. The default is supposed to be different for each axis, and even when I specified colours it didn't work. 

Indeed: https://vispy.org/api/vispy.visuals.xyz_axis.html#vispy.visuals.xyz_axis.XYZAxisVisual

Not sure about this. The only think I can think of is that we're using `gl+`, but this explicitly says `gl`:

https://github.com/sanjayankur31/vispy/blob/659369b37f86d44f98e413deb92df99631ef4d2a/vispy/visuals/xyz_axis.py#L27

I don't know enough to know if it's causing the issue. 
> So, I should drop it and change the custom way to have different colors?

Yeh, this looks simple enough, so if vispy's version doesn't do what we want, let's use our own implementation.

> 
> I'll check as well if it's possible to create our own orbit function 

I did a quick search and rotating about y axis using elevation/azimuth isn't straightforward from the looks of it. Will have to do some more digging

> However, having the camera point to the y axis does make the overall turntable camera a bit more rigid when a user wants to rotate it by hand as well for some reason. If you want to check that as well

Uh, hrm.

Yeh, we're sort of going against the "norm" for vispy here..

lej0hn:
I think I managed to do the rotation around the y axis by using the camera's roll parameter

sanjayankur31:
Uh, so, I went ahead and grepped the code, and turns out, one can set "y is up":

https://github.com/vispy/vispy/blob/659369b37f86d44f98e413deb92df99631ef4d2a/vispy/scene/cameras/base_camera.py#L194

So, if after defining the Turntable camera, all you have to do is: `cam2.up = "y"`, and it makes "y be up".. Sorry about that, should've grepped the code before instead of just searching the docs.

Here is the Hay et al cell (which is laid out along the Y axis originally) with just the one extra line:

https://github.com/NeuroML/pyNeuroML/assets/102575/972d29f4-d95b-4e86-8355-043635dc865b

Even orbiting "just works".

So, no need to look at orbiting and all that, and no need to make the camera face the Y axis manually---this one setting seems to handle all that.

sanjayankur31:
@lej0hn : do let me know when you want me to review this (we'll mark it as ready for review).

Also: ruff found a couple of minor issues, could you fix those too when you have second please? (see the "ruff" CI task for details)

lej0hn:
Ready for review

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -769,6 +871,10 @@ def plot_3D_cell_morphology(
         axon_segs = cell.get_all_segments_in_group("axon_group")
     except Exception:
         axon_segs = []
+    
+    if isPCA:
+        cell = PCA_transformation(cell)
+    current_canvas = current_view = None
 
     if current_canvas is None or current_view is None:
         view_min, view_max = get_cell_bound_box(cell)
```
sanjayankur31, line 880: I guess we should also set the view to origin etc. here as we do for `plot_interactive_3d`?

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -1069,10 +1176,14 @@ def plot_3D_schematic(
...
 
     :type color: str
+    :param isPCA: bool that if True, transforms single Cell according to its PCA to make it look 'upwards'
+    :type isPCA: bool
     """
     if title == "":
         title = f"3D schematic of segment groups from {cell.id}"
 
+    if isPCA:
+        cell = PCA_transformation(cell)
```
sanjayankur31, line 1186: do we need to clear the canvas here, and set the camera to origin etc?

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -745,6 +845,8 @@ def plot_3D_cell_morphology(
           is used)
 
     :type highlight_spec: dict
+    :param isPCA: bool that if True, transforms single Cell according to its PCA to make it look 'upwards'
+    :type isPCA: bool
```
sanjayankur31, line 849: probably worth adding the `axes_pos` argument here too, so that users can activate it when viewing single cells too (same for schematic figure)?

sanjayankur31:
One more thing---do check if the pre-commit hook is running for you. It should remove empty lines and trailing spaces etc.

Try running:

```
pre-commit install
```

in your virtual env, and then it should run on all changed files each time you commit.

lej0hn:
> Looks very good.
> 
> I was testing it with another pyramidal cell here:
> 
> https://github.com/sanjayankur31/Human-L2-3-Cortical-Microcircuit/blob/feat/neuroml/NeuroML2/HL23PYR.cell.nml
> 
> It's translated correctly, but the rotation doesn't look right. Do you want to try to see why? It's not quite "upright", but should be given the morphology? 🤔
> Screencast.from.2024-06-11.16-45-08.webm
> 
> Here's another example from the Allen cell types database:
> 
> https://github.com/OpenSourceBrain/AllenInstituteNeuroML/blob/master/CellTypesDatabase/models/NeuroML2/Cell_329321704.cell.nml
> 
> Also not quite upright:
> Screencast.from.2024-06-11.16-50-16.webm
> 
> Am I doing something wrong?

No, you were right, there was a missing '-' sign in the angles. But that reminds me, in the case of the first cell for example, do we wish to see it like this, or opposite?
Screenshot from 2024-06-11 19-58-30

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -204,12 +209,24 @@ def create_new_vispy_canvas(
             [axes_pos[0], axes_pos[1], axes_pos[2] + axes_length],
         ]
         scene.Line(
-            points,
-            connect=numpy.array([[0, 1], [0, 2], [0, 3]]),
+            [points[0], points[1]],
             parent=view.scene,
-            color=VISPY_THEME[theme]["fg"],
+            color='red',
```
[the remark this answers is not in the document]
lej0hn: very unfortunately not that I could find ..., tried it both like so and with rgb values and it didnt do the trick.

lej0hn:
> One more thing---do check if the pre-commit hook is running for you. It should remove empty lines and trailing spaces etc.
> 
> Try running:
> 
> ```
> pre-commit install
> ```
> 
> in your virtual env, and then it should run on all changed files each time you commit.

I have already installed it

sanjayankur31:
> No, you were right, there was a missing '-' sign in the angles. But that reminds me, in the case of the first cell for example, do we wish to see it like this, or opposite? Screenshot from 2024-06-11 19-58-30

This is the correct way---that long straight line is an artificial axon that the modellers have added there. The apical dendrites here are correctly going up.

I don't know how to generalise this mathematically though. We can't say "soma is always down", and we can't necessarily say "long dendrites always go up" because that may not be the case for all cells :thinking: 

Not sure how NeuroML-db have done it---I think they did it manually:

https://neuroml-db.org/gallery

sanjayankur31:
> > One more thing---do check if the pre-commit hook is running for you. It should remove empty lines and trailing spaces etc.
> > Try running:
> > ```
> > pre-commit install
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in your virtual env, and then it should run on all changed files each time you commit.
> 
> I have already installed it

Hrm, is it running when you commit? Check once, I thought I saw some trailing spaces that it should have removed.

pyneuroml/plot/PlotMorphologyVispy.py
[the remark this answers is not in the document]
sanjayankur31: I think it needs to be rgba, where "a" is just the "alpha" (transparency?) from the looks of it:

https://vispy.org/api/vispy.visuals.line.line.html#module-vispy.visuals.line.line:


> colorColor, tuple, or array
>    The color to use when drawing the line. If an array is given, it must be of shape (…, 4) and provide one rgba color per vertex. Can also be a colormap name, or appropriate Function.

lej0hn:
> > No, you were right, there was a missing '-' sign in the angles. But that reminds me, in the case of the first cell for example, do we wish to see it like this, or opposite? Screenshot from 2024-06-11 19-58-30
> 
> This is the correct way---that long straight line is an artificial axon that the modellers have added there. The apical dendrites here are correctly going up.
> 
> I don't know how to generalise this mathematically though. We can't say "soma is always down", and we can't necessarily say "long dendrites always go up" because that may not be the case for all cells 🤔
> 
> Not sure how NeuroML-db have done it---I think they did it manually:
> 
> https://neuroml-db.org/gallery

Just by playing around with models I found that this condition works  :
if the z angle of rotation < 0 then do z_angle += np,pi in order to rotate it 180 degrees. There probably are cases where it doesn't apply, but maybe it satisfies the majority of them

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -471,7 +493,10 @@ def plot_interactive_3D(
         logger.debug(f"center, view_min, max are {center}, {view_min}, {view_max}")
 
     else:
+        singleCell = True
```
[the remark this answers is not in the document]
lej0hn: Maybe we could use that to see if its a single cell, otherwise throw an exception?

pyneuroml/plot/PlotMorphologyVispy.py
lej0hn: axes_pos argument doesn't exist in plot_interactive_3d either, should I add in every one of those?

pyneuroml/plot/PlotMorphologyVispy.py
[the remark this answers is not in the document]
lej0hn: Yes, just forgot the 'a' when I wrote it :) . I used the color array from here as an example 
https://github.com/sanjayankur31/vispy/blob/659369b37f86d44f98e413deb92df99631ef4d2a/vispy/visuals/xyz_axis.py#L27
It always colors all axis or lines, as the first color of the array provided

sanjayankur31:
> Just by playing around with models I found that this condition works : if the z angle of rotation < 0 then do z_angle += np,pi in order to rotate it 180 degrees. There probably are cases where it doesn't apply, but maybe it satisfies the majority of them

Cool, let's go with that for now. Please do document it in the docstring I guess, just so users are aware that this is how we do it.

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -356,6 +373,11 @@ def plot_interactive_3D(
         If you have a good GPU, you can increase both these values to get more
         detailed visualizations
     :type precision: (int, int)
+    :param upright: bool only applicable for single cells: Makes cells "upright"
+    (along Y axis) by calculating it PCA and transforming cell co-ordinates to
```
sanjayankur31, line 377: I *think* multiline docstrings require the next line to be indented (check the rtd CI task to see what it looks like to be sure)

lej0hn: Hm, changed it but it still shows the first line only. This happens to the other parameters with multiple lines too however..

pyneuroml/plot/PlotMorphologyVispy.py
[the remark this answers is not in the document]
sanjayankur31: Ugh, that's bad---sounds like a vispy bug.. not important though, we can look into it later.

pyneuroml/plot/PlotMorphologyVispy.py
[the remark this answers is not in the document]
sanjayankur31: I *think* if it gets to this branch of the conditional, it *should* already be a single cell, otherwise the bits up till here would already error.

pyneuroml/plot/PlotMorphologyVispy.py
sanjayankur31: Ah, yeh. We can think about whether we want to also expose this in the `pynml-plotmorph` command line tool later.

pyneuroml/plot/PlotMorphologyVispy.py
sanjayankur31: OK, let's leave it for the moment. I can take a look at it after we merge and see what's going on.

sanjayankur31:
@lej0hn : please do let me know when you want me to take another look at this one.

lej0hn:
Could you take another look?

sanjayankur31:
Sure, will try it today, but we've got a divisional meeting coming up so it may have to be tomorrow during our catch up :(

sanjayankur31:
- cells/network to try: https://github.com/OpenSourceBrain/ACnet2/blob/master/neuroConstruct/generatedNeuroML2/MediumNet.net.nml

PR #380, opened 2024-06-04
Feat/loadswc  Adding SWC loader to load SWC file

AdityaBITMESRA:
This pull request adds a new module for loading and parsing SWC files. It also defines a simple data structure for holding the parsed data in memory, with segments and segment groups.

The data structure is a rough first version and will be updated as needed.


pyneuroml/swc/LoaderSWC.py
```diff
@@ -0,0 +1,110 @@
+class Point:
+    def __init__(self, x, y, z, diameter):
+        self.x = float(x)
+        self.y = float(y)
+        self.z = float(z)
+        self.diameter = float(diameter)
+
+
+class Segment:
+    def __init__(self, segment_id, name=None):
```
sanjayankur31, line 10: I was thinking about this yesterday. Instead of re-defining these classes, can we not just use the classes already defined in libNeuroML?

- `neuroml.Morphology` -> can be the "top level morphology object" either as is, or extended to `MorphologySWC` or something if it needs tweaking?
- `neuroml.Segment`
- `neuroml.SegmentGroup`
- `neuroml.Point3DWithDiam`

Finally, if we use `neuroml.Cell` as the overall object, you get to use all the helper functions to add segments/segment groups, optimise the morphology and the graph representation related functions of the morphology that we've implemented for `neuroml.Cell` too.

For example, if we simply collect all the segments and make sure they pass the basic tests (which could be added to the neuroml.Cell class), we could maybe just even use the `create_unbranched_segment_groups` helper function to automatically create the first iteration unbranched segment groups:

https://libneuroml.readthedocs.io/en/latest/userdocs/coreclasses.html#neuroml.nml.nml.Cell.create_unbranched_segment_group_branches

So, what's the advantage of redefining them vs just reusing them? What do you think?

pyneuroml/swc/LoaderSWC.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def add_segment(self, segment):
+        self.segments[segment.id] = segment
+        if segment.parent_id == -1:
+            self.root = segment
+        elif segment.parent_id in self.segments:
+            segment.parent = self.segments[segment.parent_id]
+            segment.parent.children.append(segment)
+
+    def create_segment_groups(self):
```
sanjayankur31, line 66: this is good for a start, but there are going to be a lot of more segment groups in the cell. Each unbranched section of the morphology is a segment group. All of these will be of one of the standard types: soma or dendrite or axon. So, we also need to detect all the SWC sections that form an unbranched chain and put those in a segment group. See the `add_unbranched_segment_group` helper function in the neuroml.Cell class:

https://libneuroml.readthedocs.io/en/latest/userdocs/coreclasses.html#neuroml.nml.nml.Cell.add_unbranched_segment_group

also:

https://libneuroml.readthedocs.io/en/latest/userdocs/coreclasses.html#neuroml.nml.nml.Cell.add_unbranched_segments

sanjayankur31 (changes requested):
It's coming along. Could you pull development into your feature branch again so that the diff here becomes cleaner and is limited to your changes only?

pyneuroml/swc/LoaderSWC.py
sanjayankur31: Please disregard this bit for the moment. Let's start with a representation that is completely independent of NeuroML---this way this will be a general (NeuroML independent) API for working with SWC that other tools can also use. 

So, instead, the way to go would be to create a data structure---a tree/graph where each node corresponds to a line in the swc storing the information there: co-ordinates/diameter/parent/type etc.

Once we have this, we can query it to ask "what are the parents of this node", or "what are all the nodes of type X that have more than one parent" etc.

For neuroml.Cell, I'd created a networkx representation of the points, and then that allows us to use the networkx API to loop over the graph and query it etc.

Take a look at the following functions in nml.py in the development branch of libneuroml:

- `get_adjacency_list`
- `get_graph`
- `create_unbranched_segment_groups`
- `get_morphology_root`



they use

PR #398, opened 2024-06-27
Support visualisation of morphology only files

sanjayankur31:
CC: @pgleeson


sanjayankur31:
Seems to work fine, hasn't broken anything. So I'll merge this :crossed_fingers:

Issue #359, opened 2024-05-14
vispy-viewer: make it run in jupyter notebooks using jupyter_rfb


sanjayankur31:
The vispy bits should work in jupyter notebooks/jupyterlab using jupyter_rfb. We need to test this out and make any changes necessary to get it to work

See also: https://github.com/vispy/jupyter_rfb/issues/89

sanjayankur31:
Works, will open new issue for multiple visualisations not working.

PR #399, opened 2024-07-03
feat/Make 3d visualization compatibile with jupyter notebooks

lej0hn:
Aims to solve #359 

To run in jupyter you need to get the returned canvas object and call it.
e.g.
`canvas = plot_interactive_3D(nml2_doc, plot_type='detailed', axes_pos=[0,0,0], upright=True)`
`canvas`


sanjayankur31:
Is it worth creating a `plot_interactive_3D_jupyter` function that wraps around the `plot_interactive_3D` function and calls `canvas` do you think? It'll pass the args to `plot_interactive_3D` but set the `nogui` etc. options as required.

Is the rbf package required for this at all? If yes, we'll  also need to define a new `jupyter` extra in setup.cfg that could  pull in the necessary deps?

lej0hn:
While it is possible, then shouldn't we have wrapper functions for each visualization function (`plot_3d_cell_morphology`, `plot_3d_schematic`) as well? Also I believe the arguments are identical, so I guess it comes down to which method would be preferred by a user -- to call a  different jupyter-specific function or to call canvas? I'm really not sure, although I think I tend a bit to calling canvas.

Yes, the rfb package is required and I believe is the only extra package required. I will add it to the setup.cfg

sanjayankur31:
We could have wrappers for all of them yes. I was simply thinking of this:


```
plot_interactive_3d_jupyter(*args, **kwargs):
    canvas = plot_interactive_3d(*args, **kwargs)
    canvas
```

(do check the usage of args and kwargs, I can't remember what the correct form is)


because I can't think of a way of informing users that `canvas` has to be called in jupyter notebooks---it'll have to be documented somewhere and users will have to find that documentation. With the wrapper, all they do is call a function and we take care of things for them "under the hood".

I guess another way is to simply add a new argument to all the functions: `jupyter` which will call canvas, similar to `nogui` and all that?


Is calling `canvas` all that is required? No other changes at all?

lej0hn:
Yes I see your point. So both the wrappers and the argument are valid solutions, and perhaps with the argument we won't have to return the canvas as well, so it remains more the same. 

Apparently yes, calling canvas, or using display(canvas) if in a function are all it needs

sanjayankur31:
Tasks:

- check that all features work in notebooks
- what happens if the methods are called repeatedly? (does the previous canvas get overwritten??)
- check if a windowed visualiser can be opened from notebooks: if not, no need to add an argument, just always call canvas if using a notebook; if yes, add an option to allow users to use the windowed visualiser 


Use neuroconstruct and play with its visualiser to get an idea of what features we can add to the pyneuroml visualiser

PR #401, opened 2024-07-04
add downloads URL as biosimulations.submit_simulation_archive return value

stellaprins:
Small tweak so that I can fetch results from models ran remotely on biosimulators.


sanjayankur31:
Hi @stellaprins , thanks for that. I tweaked it a bit so that it returns all the URLs, not just the download URL:


https://github.com/NeuroML/pyNeuroML/commit/81abeea27b266a9c56320ecab3ed2d5b34638ca6

So one can now do:

```
res = submit_simulation_archive(..)
download_url = res["download"]
```

and so on.

sanjayankur31:
@all-contributors please add @stellaprins for code, ideas,

PR #405, opened 2024-07-16
feature/Added lighting

lej0hn:
Shading addition on meshes in the 3d visualization


sanjayankur31:
Hrm, something weird happens with this---the soma is no longer shown. This is without lighting:

image


This is with:

image


The soma cylider is missing (at least). Even on zooming in, it isn't there, so I don't think it's a case of it being in shadow:

image

lej0hn:
You were right, it had a few problems --it seems that it was caused by the InstancedShadingFilter. After changing to ShadingFilter, I believe it looks much better

sanjayankur31:
Cool, I'll test it out later. What's the difference between the two? It seems that `InstancedMeshFilter` should be the one to use---is this a vispy bug, or is it something to do with the way we're creating our meshes? Perhaps testing it out with a simple non-neuroml example just to check what's going on would be a good thing to do?

lej0hn:
yeh probably a good idea

lej0hn:
Hey @sanjayankur31, here I'm trying to move the light as the camera moves. The values of the light_dir seem to be changing normally, yet there is not change on the visualization. Do you have any ideas?

sanjayankur31:
I'm not entirely sure tbh. I found this issue on some digging which seems to do exactly what we want. Did you run into it yet?

https://github.com/vispy/vispy/issues/1479

lej0hn:
> I'm not entirely sure tbh. I found this issue on some digging which seems to do exactly what we want. Did you run into it yet?
> 
> [vispy/vispy#1479](https://github.com/vispy/vispy/issues/1479)

Not this specific issue but it seems that the solution is what they used in the shading example that I used too. So basically it looks the same with what I did. 

https://vispy.org/gallery/scene/mesh_shading.html

pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -1176,7 +1177,25 @@ def create_instanced_meshes(meshdata, plot_type, current_view, min_width):
...
+            shading_filter.light_dir = light_dir[:3]
+            initial_light_dir = current_view.camera.transform.imap(light_dir)
+
+            @current_view.scene.transform.changed.connect
+            def on_transform_change(event):
+                transform = current_view.camera.transform
+                shading_filter.light_dir = transform.map(initial_light_dir)[:3]
+                # print(shading_filter.light_dir)
+
+    attach_headlight(current_view)
```
sanjayankur31, line 1198: this needs to be indented so that it's called for each new mesh. At the moment, it's only being called once for all meshes, which means it's only being applied for the last mesh..

sanjayankur31: That doesn't fix the issue though. I'd file an issue on the vispy repo and leave this for the moment. Digging into vispy internals is not the best use of our time.

sanjayankur31:
I've copied your code over and added it now in https://github.com/NeuroML/pyNeuroML/commit/562c8249516ad8b15307a6f0e4345c6edaf62a4b

It works fairly well. Getting the direction of the light right was a bit tricky, but I came up with some simple heuristics depending on the boundaries of the view box. Will do for the moment.

PR #407, opened 2024-07-18
Feat/export swc:- This aims to add a file which exports SWC file into neuroml. 

AdityaBITMESRA:
**This draft PR aims to add a file which exports SWC into neuroml for visualisation**

1. It is not complete mainly around the soma representation part.
2. Will be making it more efficient and reliable so that it can cover all swc cases.


AdityaBITMESRA:
Hi @sanjayankur31 I have refactored the code. Can you give it a look if time permits?

sanjayankur31:
Did we add any unit tests for the writer @AdityaBITMESRA ? (Not seeing them in this PR)

AdityaBITMESRA:
No, I will add them soon

AdityaBITMESRA:
I wanted to ask how should I structure my test file like I should build a mock swc data for all the cases?

sanjayankur31:
> I wanted to ask how should I structure my test file like I should build a mock swc data for all the cases?

Yeh, you can build mock data, or you can even copy the case files from the CVApp if you want, or you can only paste their contents into a Python string in the test? Whatever you think works best.

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
+from typing import Dict, List, Optional, Set
+
+from LoadSWC import SWCGraph, SWCNode
```
sanjayankur31, line 3: use a relative import here: `from .LoadSWC ..`, that will tell Python to look for `LoadSWC` in the same folder as `ExportNML`.

https://docs.python.org/3/reference/import.html#package-relative-imports

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+
+
+class NeuroMLWriter:
+    def __init__(self, swc_graph: "SWCGraph"):
```
sanjayankur31, line 7: does `SWCGraph` need to be in quotes here?

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+            or self.section_types[1].lower() != "soma"
+        ):
+            print("Error: null data or section types in nmlWrite")
+            return ""
+
+        cell_name = self.get_cell_name()
+        start_point = self.find_start_point()
+
+        print(f"Cell name: {cell_name}")
+        print(f"Start point: {start_point}")
```
sanjayankur31, line 53: if you're using these for debugging, use `logging` so that you can use `logger.debug` and so on. Only stuff that is definitely needed to be printed to the user should use `print`.

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+        self,
+        this_point: SWCNode,
+        parent_point: SWCNode,
+        cable_id: int,
+        version: str,
+        new_cell: bool,
+    ):
+        """
+        Handle the soma point in the neural structure.
+
```
sanjayankur31, line 186: It'll be good to include an explanation of the different soma cases that we have to handle, with a link to the pdf file. The idea is that someone else reading the code should be able to see the reasoning behind the function (without having to go looking to figure it out themselves). So write a nice long docstring here that explains the multiple cases---it'll also help you check your own understanding of the logic you've used.

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+        version: str,
+        new_cable: bool,
+    ):
+        """
+        Create a segment in the NeuroML structure.
+
+        :param this_point: The current SWCNode.
+        :param parent_point: The parent SWCNode.
+        :param cable_id: The ID of the current cable.
+        :param version: The NeuroML version being used.
```
sanjayankur31, line 263: Sorry, we should've mentioned this---NeuroMLv1 is deprecated and not to be used at all. So, no need to have special cases supporting it. We only export to NeuroMLv2.

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+        :param new_cable: Whether this point starts a new cable.
+        """
+        print(
+            f"Creating segment: Point {this_point.id}, Parent {parent_point.id}, Cable {cable_id}"
+        )
+        seg_id = self.next_segment_id
+        self.next_segment_id += 1
+        self.point_indices_vs_seg_ids[this_point.id] = seg_id
+
+        segment = [f'      <segment id="{seg_id}" name="Seg_{seg_id}"']
```
sanjayankur31, line 273: Now that this is all neuroml specific, is there a reason we're not creating objects using the libneuroml API? (`Segment` and `SegmentGroup`)? It's usually better than creating xml by hand.

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+            SWCNode.AXON: ["axon_group", "color_grey"],
+            SWCNode.BASAL_DENDRITE: ["basal_dendrite", "dendrite_group", "color_green"],
+            SWCNode.APICAL_DENDRITE: [
+                "apical_dendrite",
+                "dendrite_group",
+                "color_magenta",
+            ],
+        }
+        groups.extend(
+            type_groups.get(p.type, [f"SWC_group_{p.type}", "dendrite_group"])
```
sanjayankur31, line 437: what are we doing here? if a group for the type is not found, we're using `SWC_group_{type}` and `dendrite_group`? Not sure if this is correct. If it doesn't belong to the standard types, it shouldn't be valid SWC :thinking:

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,501 @@
...
+                else ""
+            )
+            self.group_content.append(
+                f'      <cable id="{cable_id}" name="{self.segment_name(this_type)}_{self.seg_per_typ[this_type]}"{fract_info}>'
+            )
+            for group in self.get_groups_for_type(this_point):
+                self.group_content.append(f"        <meta:group>{group}</meta:group>")
+            self.group_content.append("      </cable>")
+        else:
+            for group in self.get_groups_for_type(this_point):
```
sanjayankur31, line 393: if this is the first point of a new branch, it should also create a new segment group. 

ALL "cables" also have to have their own segment groups, in addition to the standard soma/dendrite etc groups. Is this being done?

sanjayankur31 (changes requested):
Looks very good. Getting there. A few notes/queries/suggestions there.

AdityaBITMESRA:
Thankyou for all the comments @sanjayankur31 I will update after addressing the comments

pyneuroml/swc/ExportNML.py
AdityaBITMESRA: Can I get some more ideas on how to proceed with this? I am bit confused about using the libneuroml API.

sanjayankur31: Sure, see this example for a start:

https://docs.neuroml.org/Userdocs/MultiCompartmentOLMexample.html#declaring-the-cell

When users use NeuroML, they don't write XML---that's tedious and error prone. Instead, they use the libNeuroML API:

```
cell = neuroml.Cell(...)
cell.add_segment(...)

..
```

it's basically a Python object model of the NeuroML schema, so a Cell contains Morphology, which contains Segments and SegmentGroups.

That's a better way of generating the NeuroML than writing XML by hand.

Let me know if this isn't clear and we can have a chat where I can explain it to you more.

AdityaBITMESRA: Thankyou I understand it now...will try to implement this

sanjayankur31: Do take a look at the various "utility functions" in the Cell class, they should help with lots of things (hopefully):

https://libneuroml.readthedocs.io/en/latest/userdocs/coreclasses.html#cell

AdityaBITMESRA:
Hi @sanjayankur31 I have tried to address all the comments so can you have a look if time permits...it is still not complete but I wanted to know if  I am going right or not

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,550 @@
...
+        """
+        Create a Cell object.
+
+        :return: The created Cell object.
+        :rtype: Cell
+        """
+        logger.info("Creating Cell object")
+        cell_name = self.get_cell_name()
+
+        self.cell = Cell(id=cell_name)
```
sanjayankur31, line 71: I'd maybe use the `component_factory` here:

```
from neuroml.utils import component_factory
cell = component_factory("Cell", id="cell_name", notes="something")
```

(the advantage of the component factory is that it validates things as we go)

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,550 @@
...
+            self.cell.morphology.segment_groups.append(cable_group)
+
+        # Add the cell_type property
+        self.cell.properties.append(
+            Property(tag="cell_type", value="converted_from_swc")
+        )
+
+        logger.info("Segment groups created successfully")
+
+    def print_soma_segments(self) -> None:
```
sanjayankur31, line 515: is this for debugging?

you can use `cell.get_segment_group("soma_group")` to get all segments in the segment group.

See `help(neuroml.Cell)` in a python interpreter to see all the methods of the `Cell` class.

sanjayankur31 (commented):
It's looking good and you're on the right track. Well done. :+1: 

Dropped a few comments about things I wasn't completely clear on.

pyneuroml/swc/ExportNML.py
AdityaBITMESRA: yes I added this for debugging...will go through the recomendation

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,550 @@
...
+            self.next_cable_id += 1
+            self.cable_ids_vs_indices[cable_id] = []
+            logger.debug(f"New cable created: Cable ID {cable_id}")
+        else:
+            cable_id = max(self.cable_ids_vs_indices.keys())
+            logger.debug(f"Using existing cable: Cable ID {cable_id}")
+
+        if this_point.type == SWCNode.SOMA:
+            self.handle_soma(this_point, parent_point, cable_id, version, new_cell)
+        elif this_point.type != SWCNode.SOMA and parent_point.type == SWCNode.SOMA:
```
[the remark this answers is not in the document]
AdityaBITMESRA: Thankyou for pointing this out...I will add comments to make it more clear

pyneuroml/swc/ExportNML.py
sanjayankur31: Cool, I'd also suggest writing unit tests for testing that each part of your code is doing what you expect it to do :)

pyneuroml/swc/ExportNML.py
```diff
@@ -0,0 +1,535 @@
+import logging
+import tempfile
+from typing import Dict, List, Optional, Set
+
+import neuroml.writers as writers
```
sanjayankur31, line 5: This looks good, well done. I think the file should be `test_ExportNML.py`, just for consistency, though.

AdityaBITMESRA:
Hi @sanjayankur31  I have addressed the issues...Can you give it a look?

AdityaBITMESRA:
I was running with this [swc file](https://github.com/NeuroML/Cvapp-NeuroMorpho.org/blob/master/examples/dCH-cobalt.CNG.swc)

AdityaBITMESRA:
This is the plotted morphology from the python based converter 



This is from the cvapp

AdityaBITMESRA:
I went through the neuroml file in the soma segment 70 the proximal is set to the previous segment distal which is creating the problem.

When I am manually writing the proximal to that of point id 1 which is its parent according to the swc data I am getting the exact morphology like the cvapp

sanjayankur31:
Cool, I think I know what's going on. I'll look into it soon.

sanjayankur31:
Working on this here: https://github.com/NeuroML/pyNeuroML/pull/421

(This will be automatically merged when that one gets merged, so leaving this open)

PR #408, opened 2024-07-18
Feat/upright 2d plotter: enable upright for 2d morph plots also

sanjayankur31:


sanjayankur31:
hrm, not sure if making cells upright is required for the 2D plotter. Closing this for the moment

PR #412, opened 2024-08-04
Test multiple shading filters

lej0hn:
Test looping through shading filters (one for each mesh)

### DIFF
--- a/pyneuroml/plot/PlotMorphologyVispy.py
+++ b/pyneuroml/plot/PlotMorphologyVispy.py
@@ -20,9 +20,10 @@ import progressbar
 from neuroml import Cell, Morphology, NeuroMLDocument, SegmentGroup
 from neuroml.neuro_lex_ids import neuro_lex_ids
 from scipy.spatial.transform import Rotation
+from vispy.visuals.filters import InstancedShadingFilter, ShadingFilter
 
 from pyneuroml.pynml import read_neuroml2_file
-from pyneuroml.utils import extract_position_info, make_cell_upright
+from pyneuroml.utils import extract_position_info, rotate_cell, translate_cell_to_coords
 from pyneuroml.utils.plot import (
     DEFAULTS,
     get_cell_bound_box,
@@ -33,15 +34,14 @@ from pyneuroml.utils.plot import (
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-pynml_in_jupyter = False
-
 try:
     from vispy import app, scene, use
-    from vispy.geometry.generation import create_sphere
+    from vispy.geometry.generation import create_cylinder, create_sphere
     from vispy.geometry.meshdata import MeshData
     from vispy.scene.visuals import InstancedMesh
     from vispy.util.transforms import rotate
 
+    pynml_in_jupyter = False
     if app.Application.is_interactive(app):
         pynml_in_jupyter = True
         from IPython.display import display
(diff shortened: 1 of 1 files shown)


pyneuroml/plot/PlotMorphologyVispy.py
```diff
@@ -1176,7 +1237,27 @@ def create_instanced_meshes(meshdata, plot_type, current_view, min_width):
             parent=current_view.scene,
         )
         # TODO: add a shading filter for light?
+        shaders.append(
+            ShadingFilter("smooth", shininess=1, ambient_light=(0.8, 0.8, 0.8, 0.6))
+        )
+        mesh.attach(shaders[counter])
         assert mesh is not None
+
+        def attach_headlight(current_view):
```
sanjayankur31, line 1246: now that you're saving the shaders, shouldn't all of this be done only once outside the for loop?

sanjayankur31 (changes requested):
So, I'm not sure this works at all (let's forget about the performance). Try this very simple cell: 

https://github.com/OpenSourceBrain/ACnet2/blob/master/neuroConstruct/generatedNeuroML2/bask_soma.cell.nml

The light does not always stay in the same direction, no?


https://github.com/user-attachments/assets/2891a386-dae9-48d5-9d3a-65a21029df31

Is it working for you?

sanjayankur31:
I'm not confident I understand what's happening here---what does map do, and what does imap do? How does one set the initial light dir to be coming from the camera's direction?

pyneuroml/plot/PlotMorphologyVispy.py
lej0hn: Probably yes you are right

lej0hn:
It does, in fact not seem, to be working. However I have tested the same thing with both meshes and instancedMeshes and Shading and InstancedShading and in the simple examples the for loop seems to get the job done. I just tested however how the InstancedMeshes do with the ShadingFilter (instead of InstancedShadingFilter, like what we are using) and that seems to create a problem since the shading gets all wrong --so my best guess is some incompatibility between those if that makes sense?

sanjayankur31:
I don't think InstancedMeshes and the ShadingFilter mix, so let's not try that any more. Do multiple InstancedMeshes work with individual InstancedShadingFilters? If they do, there's no reason they shouldn't work in a for loop. So, confirming that is our first step. If they work in the simple examples, they should also work for our meshes---no reason why they shouldn't. What are we doing different to the examples?

lej0hn:
> I don't think InstancedMeshes and the ShadingFilter mix, so let's not try that any more. Do multiple InstancedMeshes work with individual InstancedShadingFilters? If they do, there's no reason they shouldn't work in a for loop. So, confirming that is our first step. If they work in the simple examples, they should also work for our meshes---no reason why they shouldn't. What are we doing different to the examples?

Multiple InstancedMeshes do work with multiple InstancedShadingFilters in the simple examples. Our problem is that when we use InstancedShadingFilter instead of the ShadingFilter, we have a big part of the cell disappear for some reason

sanjayankur31:
> > I don't think InstancedMeshes and the ShadingFilter mix, so let's not try that any more. Do multiple InstancedMeshes work with individual InstancedShadingFilters? If they do, there's no reason they shouldn't work in a for loop. So, confirming that is our first step. If they work in the simple examples, they should also work for our meshes---no reason why they shouldn't. What are we doing different to the examples?
> 
> Multiple InstancedMeshes do work with multiple InstancedShadingFilters in the simple examples. Our problem is that when we use InstancedShadingFilter instead of the ShadingFilter, we have a big part of the cell disappear for some reason

Yeh, but that means that it doesn't quite work as it should. We need to understand why. Is it an issue with how we're doing it, or is it a vispy issue (and if it is, we need to report it upstream so they can help fix it).

How about you test with some very simple cells that have fewer than 10 segments of different shapes, like the one cylinder example I've given above. That'll help you narrow down where the issue is. If it works with 10 segments, make a bigger cell---20, then 40, and so on. 

You can either take a large cell and remove lots of segments from it to make it smaller, or you can write a simple script to create cells. See this example for the latter, it uses a few utility functions to quickly create a very simple cell:

https://docs.neuroml.org/Userdocs/MultiCompartmentOLMexample.html#declaring-the-cell

lej0hn:
Right, since that is a distinct issue however maybe we can close the one open in vispy, since the answer seems to cover our needs for now. Also InstancedShadingFilter doesn't seem to be working with the one cylinder, which I guess is the simplest form. Since the most basic cell doesn't work, do you have any ideas on how to go about debugging it? (to understand whether the fault is in our implementation or vispy's)

sanjayankur31:
Since we're not using one mesh, we're not using InstancedMeshes any more, and so not using InstancedShadingFilters either. I picked bits from here as part of https://github.com/NeuroML/pyNeuroML/commit/562c8249516ad8b15307a6f0e4345c6edaf62a4b

PR #413, opened 2024-08-06
Improve checks to ensure that the right input files are provided for each pynml command

sanjayankur31:


sanjayankur31:
@stellaprins @robertvi: could you also please take a quick look to check that I haven't messed up the sbml/sedml bits here?

sanjayankur31:
CC: @subhacom : we'd not checked all the cases, this should fix that.

PR #420, opened 2024-08-21
Add segment picking ability in 3D

lej0hn:
- Ability to click on any segment and get the InstancedMesh that was clicked
- Get the instanced_position of said mesh
- Get info about the corresponding segment*

-TODO
* Fix type of dictionary keys of positions-segment_infos because the value is right but it throws error
(to check that error uncomment lines 1219-1220)
* Add necessary parameters/arguments to plot_schematic_3d

### DIFF
--- a/pyneuroml/plot/PlotMorphologyVispy.py
+++ b/pyneuroml/plot/PlotMorphologyVispy.py
@@ -727,7 +727,7 @@ def plot_interactive_3D(
                     except KeyError:
                         pass
 
-                    plot_3D_cell_morphology(
+                    meshdata, segment_info = plot_3D_cell_morphology(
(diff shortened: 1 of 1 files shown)


sanjayankur31:
Just tested this out---seems to work well, well done! :sparkles: 

To begin with, could you maybe please comment the new functions to explain the logic in use there? It'll just make it easier for others to understand how this is all working. So, nice long docstrings that explain the working of the functions would be good.

sanjayankur31:
Thanks---it'll also make it easier for us to review :)

sanjayankur31:
Replaced by https://github.com/NeuroML/pyNeuroML/pull/428

PR #428, opened 2024-09-13
Added segment picking from faces 

lej0hn:
- Used FacePickingFilter to get face clicked
- Get segment from face 
- Get info for segment

- Is disabled for schematic viewing


sanjayankur31:
Works very well @lej0hn :+1:  . Could you link to the sources/docs you used to come up with this? I'll comment the code so that its easier to update/modify/maintain in the future for us.

I've got it merged in my local branch, but will push once I've commented the code etc.

lej0hn:
> Works very well @lej0hn 👍 . Could you link to the sources/docs you used to come up with this? I'll comment the code so that its easier to update/modify/maintain in the future for us.
> 
> I've got it merged in my local branch, but will push once I've commented the code etc.


Great, if you need anything regarding the docs please let me know!
The main resource is regarding the face_picking_filter that the implementation was based on which is in this link:
https://vispy.org/gallery/scene/face_picking.html#sphx-glr-gallery-scene-face-picking-py

sanjayankur31:
Cool---does the example change the color (or use some other method) to indicate what was clicked on perhaps? I'm trying to figure out how to pull this in in its current form. Perhaps a quick QT wrapper app that has an "info" pane to just print what we have so far would be a start. What do you think?

PR #430, opened 2024-10-02
Feat: util to create + include new components in NeuroMLDocuments

sanjayankur31:
Adds a utility function to allow users to add new components of user-defined component types to their models.
It creates the component using pylems and saves it to a new xml file. It then includes the xml file in the main model document.

### DIFF
--- a/pyneuroml/utils/components.py
+++ b/pyneuroml/utils/components.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python3
+"""
+Utilities for creation and inclusion of new Components in models.
+
+File: pyneuroml/utils/components.py
+
+Copyright 2024 NeuroML contributors
+"""
+
+import logging
+from typing import Optional
+
+from lems.model.component import Component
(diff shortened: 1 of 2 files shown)


sanjayankur31:
@pgleeson @borismarin : what do you think of this? 

Traditionally, the way we'd do this is to edit the XML file to manually add the new component to the model. Or, we'd define the new component with the new component type. This just does that programmatically---it allows me to create a new component with new parameters using python and saves it to a new XML file that is included in my model.

So, I no longer need to hard code the component with the component type, nor do I need to run my python to create the model and then go edit the XML manually.

sanjayankur31:
PS: not added it as a function of `NeuroMLDocument` in libNeuroML to clearly separate what is part of the standard and what is extending it.

sanjayankur31:
Updated: returns the name of the new file, but the user still has to manually include this in the LEMS simulation file. Components are not part of the NeuroML standard and as LEMS bits, they cannot be included in NeuroML files.

PR #436, opened 2024-11-06
Fix file compatibility check

### LINKED ISSUE #435: Improve model standard detection from files
See: https://github.com/OpenSourceBrain/SBMLShowcase/issues/77

At the moment, we detect files using their file extensions, but this isn't perfect because people may just use "xml" for files, in which case our tests will fail.

So, what we should do is:

- check the extension
- if that fails, for example, if it's "xml" which is ambigious, we read the first and last N lines of the file to grep them for tags

sanjayankur31:


sanjayankur31:
CC: @stellaprins 

I *think* this should catch everything now.

pgleeson:
@sanjayankur31 Looks good. At the end of the day nml & lems toolchain should accpet any suffix for valid nml/lems withoutt failing, but I guess these will only raise warnings, unless you've asked them to fail on mismatch

sanjayankur31:
Yeh, it checks the extension first, if that does not match what we expect it parses the xml file to figure out the root node of the xml doc. If the root node does not match, it will fail. Should it only warn even then?

pgleeson:
Does this PR also allow *.xml files to be validated as NeuroML?.
Currently breaking modelspec checks:
https://github.com/ModECI/modelspec/actions/runs/12282826198/job/34274894800

sanjayankur31:
Yes, it will. I'll merge this and cut a release so that it gets used by modelspec etc.

paritoshparashar:
Hello, I was reading the codebase and wanted to contribute.
I was wondering if the 'confirm_file_type' function can be made more safe by handling some extra errors
Like - 
1: If there is a problem in opening the file using open(filename)
2: If the filename consists of more than one '.' dots, the current splitting function would not perform efficiently there.

sanjayankur31:
Hi @paritoshparashar , thanks for your interest. I think this issue has been fixed already. (The related issue remains open because the release is made from the development branch and not main)

PR #437, opened 2024-11-06
wip: improve `pynml-summary`

sanjayankur31:
Adds an intermediate `-v`, but needs more work to properly replicate earlier behaviour.


sanjayankur31:
Needs some thought, since it also involves libNeuroML.

sanjayankur31:
Should now work with all NeuroMLDocuments, also including channels and so on---it just defaults to `whatever.info()` if there's no special handling.

Issue #433, opened 2024-10-24
Add utility function to enumerate all model parameters


sanjayankur31:
Something like `pynml-listparams <nml file>` should list all the parameters in the ComponentType definition.

Related: `pynml-showmetadata`: should print out any metadata that has been included in rdf annotations.

samkitshah1262:
Hi @sanjayankur31 , I'd like to work on this.

sanjayankur31:
Hi @samkitshah1262 , sure! I'll assign this to you. 

The general idea would be to load a file into a NeuroMLDocument object, iterate over  the objects contained in it, get their members (using the `info` method, for example), and then to print these out nicely in a readable form. 

https://libneuroml.readthedocs.io/en/latest/userdocs/Other_list.html#neuroml.nml.generatedssupersuper.GeneratedsSuperSuper.info (this is inherited by all component type classes)

Daman2461:
Hello sir, Is this issue still open? If so, I would like to work on this.

samkitshah1262:
Hi @Daman2461, I am working on this issue, couldn't work on it last month due to holidays so might look stale. Could you please pick another. Thanks!

PR #439, opened 2025-01-16
[WIP] utility function list params in ComponentType

samkitshah1262:
towards #433 , @sanjayankur31 , could you please take a look and provide feedback. Thank you !


sanjayankur31:
@samkitshah1262 : could you add a unit test too please? It just makes it easier to test + verify?

samkitshah1262:
ready for review

sanjayankur31:
Hi @samkitshah1262 , thanks for this and sorry for the delay. it's a good start, but it's incomplete because every NeuroML class is actually a "ComponentType":

https://docs.neuroml.org/Userdocs/LEMS_elements/Definingcomponenttypes.html

and example:

https://github.com/NeuroML/NeuroML2/blob/master/NeuroML2CoreTypes/Cells.xml#L1298

So, one has to load the model in LEMS (using PyLEMS) and then in the LEMS representation, go through all the component types to list their parameters

https://github.com/LEMS/pylems/

(I do want to note that we're not putting forward a GSoC project this year related to developing NeuroML software like we did last year, so if you're looking for GSoC projects, you may want to look at the other options).

samkitshah1262:
@sanjayankur31 Thank you for the review and the heads up! Was looking for GSOC project primarily, nevertheless I would love to contribute! Moreover, is there a collaboration space (except gitter) , If not a dedicated contributor channel for communication?

sanjayankur31:
> @sanjayankur31 Thank you for the review and the heads up! Was looking for GSOC project primarily, nevertheless I would love to contribute! Moreover, is there a collaboration space (except gitter) , If not a dedicated contributor channel for communication?

Most of our collaboration happens here over GitHub, and the element/matrix (which used to be gitter) channel is where folks chat if something requires synchronous communication.

PR #441, opened 2025-02-10
fix(morph-plot): handle files with single point cell types

sanjayankur31:
If a file contains the definition of a point cell type, for example, `IafTauRefCell`, we should still be able to plot the file, even if its only a single circle/sphere. Until now, we only handled files if they had multi-compartmental `Cell` or `Morphology` instances, and the plotter would crash for files that had other point cell types.


sanjayankur31:
TODO: add a simple regression test

PR #445, opened 2025-04-28
Feat test py313

sanjayankur31:


sanjayankur31:
Won't work yet, NEURON doesn't provide a wheel for 3.13.

sanjayankur31:
Neuron and pytables both provide wheels for py313 now, testing to see if that's enough.

PR #446, opened 2025-05-09
feat: remove numpy version pin

sanjayankur31:
Pytables now supports numpy>2.x

https://github.com/PyTables/PyTables/releases/tag/v3.10.1


pgleeson:
FYI pin on numpy in libnml is propagating to here and making this unpin irrelevant

PR #450, opened 2025-06-12
Feat: provide `jnml` command that will use the bundled jar

sanjayankur31:
Works fine, but it'll be good to see if it breaks any workflows---omv for example should just find this wrapper now and work with it, but we should just test it out a little.


sanjayankur31:
This seems to work fine, along with https://github.com/OpenSourceBrain/osb-model-validation/pull/138

Should I merge and cut a release @pgleeson ?

PR #453, opened 2025-07-02
ci: remove numpy pin

sanjayankur31:


sanjayankur31:
We removed this in our requirements, but forgot to remove it here from the looks of it.

PR #457, opened 2025-07-31
Refactor and improve plotting of time series and spikes from simulation generated data

sanjayankur31:


sanjayankur31:
@pgleeson : mostly refactoring and some tweaks. I've made sure that the `reload_saved_data` function continues to function as before, but it's worth checking in some places where it's being used just to be sure.

(The one change is that multiple plots open in separate plots, rather than in a multi-plot.)

pgleeson:
Ok, testing in experimental branch...

sanjayankur31:
@pgleeson : i made more commits here, so I've also updated the experimental branch to include them now. Please `git pull` when you use it.

PR #466, opened 2026-04-22
feat: add script to generate python scripts from xml

sanjayankur31:
This will be useful in multiple cases:

a) creating python scripts from xml models as starting templates for users
b) for llms/agents---where we want them to write python, not xml


sanjayankur31:
@HengyeZhu: would you have time to review this please?

HengyeZhu:
No problem! I’ll start working on it soon.

HengyeZhu:
Hi @sanjayankur31,

Sorry for the delayed reply. This PR may be fine, except for inline-cell cases such as https://github.com/OpenSourceBrain/262670/blob/master/NeuroML2/KC.net.nml.

```python
from pyneuroml import pynml

def test_inline_cell_validate():
    model = read_neuroml2_file("HH_example_net.nml", include_includes=True)

    with tempfile.NamedTemporaryFile("w", dir=".") as f:
        write_neuroml2_file(model, f.name)

        script = NmlPythonizer(f.name, "temp_nml2py_1").write()[-1]
        exec(Path(script).read_text(), {})

        assert pynml.validate_neuroml2(Path(f.name).name)
```

With the current code, this fails because the generated Python includes the extracted inline cell as:

```python
nml_doc.add("IncludeType", href='hh_cell.cell.nml')
```

while the extracted inline cell file is written to:

```text
temp_nml2py_1/hh_cell.cell.nml
```

After generating the extracted-component href relative to the output directory, the test passes.

pyneuroml/utils/nml2py.py
```diff
@@ -0,0 +1,540 @@
...
+    """
+
+    def __init__(self, nml_file: str, output_dir: str | None = None) -> None:
+        self.nml_file = nml_file
+        self.nml_file_path = Path(nml_file)
+        self.output_dir = (
+            (self.nml_file_path.parent / Path(output_dir).stem)
+            if output_dir
+            else (Path(nml_file).parent / Path("output_nml2py"))
+        )
```
HengyeZhu:
```python
self.output_dir_name = Path(output_dir).stem if output_dir else "output_nml2py"
self.output_dir = self.nml_file_path.parent / self.output_dir_name
```

pyneuroml/utils/nml2py.py
```diff
@@ -0,0 +1,540 @@
...
+
+        # Original includes (preserved as-is)
+        for href in self._original_includes:
+            # we do not move original includes, so we must tweak the path to
+            # refer to the parent directory
+            lines.append(f'nml_doc.add("IncludeType", href={href!r})')
+
+        # Extracted inline components
+        for comp, filename in self._extracted:
+            lines.append(f'nml_doc.add("IncludeType", href={filename!r})')
```
HengyeZhu, line 353:
```python
# Extracted inline components are written to the output directory.
for comp, filename in self._extracted:
    href = Path(self.output_dir_name, filename).as_posix()
    lines.append(f'nml_doc.add("IncludeType", href={href!r})')
```

sanjayankur31:
> Hi @sanjayankur31,
> 
> Sorry for the delayed reply. This PR may be fine, except for inline-cell cases such as https://github.com/OpenSourceBrain/262670/blob/master/NeuroML2/KC.net.nml.
> 
> ```python
> from pyneuroml import pynml
> 
> def test_inline_cell_validate():
>     model = read_neuroml2_file("HH_example_net.nml", include_includes=True)
> 
>     with tempfile.NamedTemporaryFile("w", dir=".") as f:
>         write_neuroml2_file(model, f.name)
> 
>         script = NmlPythonizer(f.name, "temp_nml2py_1").write()[-1]
>         exec(Path(script).read_text(), {})
> ```

Shouldn't the script be run after changing into the `temp_nml2py_1` directory? 

The idea is that all the new files are generated in a new folder and must be run there. A few reasons for this:

- keep the original model files completely separate from the new generated files
- do not risk overwriting original files (if the `id` etc. matches in your test, for example, the `HH_example_net.nml` file could be overwritten)
- simplifying paths: we could technically add folders/relative paths to the include hrefs, but that then means that whenever these extracted files are to be used, the same folder structure will have to be followed. That's not a bad thing, necessarily, but i thought it's better to use a flat directory structure, and then users can move files around and update paths as they wish.

What do you think?

HengyeZhu:
> The idea is that all the new files are generated in a new folder and must be run there.

I think this is completely fine! I realize why I could not run https://github.com/OpenSourceBrain/262670/blob/master/NeuroML2/KC.net.nml: I was launching it directly from the CLI without setting `include_includes=True`, so the include paths were wrong in that case.

sanjayankur31:
Cool, do you think it's OK to merge the PR?

HengyeZhu (approved):
No problem!

sanjayankur31:
@all-contributors please add @HengyeZhu for code, design, review

